PR #48, opened 2026-05-23
feat: CLI mints scoped tokens for release; per-project scope on OCI push (ALIEN-139)

ItamarZand88:
## Summary

Two OSS-side changes that together let `alien release` work against a manager when the caller is authenticated with an OAuth bearer, plus follow-up CLI hardening:

1. **CLI** exchanges the user OAuth JWT for a short-lived, project-and-manager-scoped token before talking to any manager endpoint, and uses that token for the artifact-registry setup + OCI push.
2. **alien-manager** gains a per-project `scopes` field on `Subject` and enforces it on the OCI push path — a token whose scopes don't cover the target project's ID is rejected at `/v2/.../blobs/uploads/`.
3. **CLI follow-ups** (a) surface the manager's response body in error messages so the user sees the actual reason (auth failure, scope mismatch, etc.) instead of just an HTTP status, and (b) unblock `alien commands invoke` in platform mode (previously errored out before reaching the manager).

OSS standalone behaviour is preserved bit-for-bit: validators that don't populate `scopes` leave it empty, and the new scope check no-ops on empty scopes — so existing single-tenant deployments still work without any token-format changes.

## Commits

| | |
|---|---|
| `feat(manager): per-project scope on Subject, enforced on OCI push` | `Subject.scopes: Vec<String>` field; `require_push_auth` suffix-matches `/<project_id>` against the JWT's scope list when non-empty. Also fixes `LocalArtifactRegistry::upstream_repository_prefix()` which was returning a hardcoded `"artifacts/default"` instead of the binding name — that was masked before because no per-project authz extracted the real project_id from the routing table. |
| `feat(cli): exchange OAuth JWT for manager-scoped token on release` | `mint_registry_push_token` + `get_or_mint_registry_push_token` in `crates/alien-cli/src/auth.rs`. Cached in keyring under `MANAGER_TOKEN_USER` (separate slot from the OAuth tokens so `logout` cleans both). `execution_context.rs` mints once at the top of the platform-mode release flow and reuses the token for `/v1/artifact-registry/*` repo setup + `/v2/...` OCI push. |
| `test: cover the per-project scope check on the OCI push path` | Extracts `scopes_cover_project` from `require_push_auth` so the security-critical predicate is unit-testable without standing up a 17-field `AppState`. 5 tests cover empty/matching/non-matching/multi-scope/leading-slash-substring-guard cases, plus 7 tests for `token_matches_target` covering match, manager mismatch, project mismatch, missing claim, malformed JWT, etc. |
| `refactor(cli): match codebase conventions on URL encoding and cache writes` | (1) `urlencoding::encode(manager_id)` in the mint URL — matches the established pattern in `execution_context.rs::resolve_url` and the gcp/aws clients. (2) `store_manager_token` is now best-effort (`warn!` on keyring write failure instead of aborting the release) — matches the worked example in `alien/CLAUDE.md` under "When `warn!` Is Appropriate". |
| `fix(cli): surface manager response body and unblock commands invoke in platform mode` ✨ NEW | (1) `create_or_get_artifact_repo` reads each response's body once via `.text().unwrap_or_default()`; surfaces the manager's structured error in the final `Failed to create artifact repository` message instead of dropping it. (2) `commands_task` branches: in platform mode, resolves the project + manager to get an authenticated client + bearer (previously `server_sdk_client()`'s platform arm errored out, blocking step 5 of onboarding); standalone/dev keep the existing path. |

## Verification

- `cargo test -p alien-manager` — **108 tests passing** across 6 binaries (was 103 before the scope-check predicate tests landed)
- `cargo test -p alien-manager --test registry_proxy_test` — 7 tests including push-auth, all green
- `cargo test -p alien-cli --features platform --lib oauth_flow::tests` — 7 tests for `token_matches_target`, all green
- `cargo build -p alien-cli --features platform` — clean
- Live end-to-end: `alien release --experimental` from a worker-template project succeeds and produces a release ID

## Backward compatibility

- OSS standalone managers run unchanged. Validators that don't set `Subject.scopes` leave it empty, and the new scope check is gated on `!subject.scopes.is_empty()`.
- The CLI's token-mint flow only activates in `#[cfg(feature = "platform")]` `ExecutionMode::Platform`. Standalone CLI use (`ALIEN_MANAGER_URL=...`) is unchanged.
- The platform-mode branch in `commands_task` is gated under `#[cfg(feature = "platform")]`; non-platform builds still go through `server_sdk_client()` directly.

## Out of scope (separate tickets)

- `Subject.scopes: Vec<String>` is loosely typed — format `"<workspace_id>/<project_id>"` enforced only by doc comment. A typed `permits_project_push(project_id)` helper on `Subject` would kill the open-coded `ends_with` matching at call sites.
- The platform-mode `alien commands invoke` flow currently re-uses the registry-push token. A separate, more-specific audience would match the operation's intent better — track in a follow-up using the documentation skill on the platform side.

## Test plan

- [ ] Reviewer runs `cargo test -p alien-manager` and confirms 108/108.
- [ ] Reviewer runs `cargo build -p alien-cli --features platform` and confirms clean.
- [ ] Reviewer confirms `Subject.scopes` default-empty behaviour preserves OSS standalone: grep for `Subject {` literals in OSS validators shows `scopes: Vec::new()` everywhere.

### DIFF
--- a/crates/alien-bindings/src/providers/artifact_registry/local.rs
+++ b/crates/alien-bindings/src/providers/artifact_registry/local.rs
@@ -147,11 +147,15 @@ impl ArtifactRegistry for LocalArtifactRegistry {
     }
 
     fn upstream_repository_prefix(&self) -> String {
-        // The embedded local registry accepts two-segment repo paths (e.g.,
-        // "namespace/repo"). We use "artifacts/default" as the canonical prefix
-        // — this matches what the CLI hardcodes in dev mode and what the proxy
-        // routing table uses to route pushes to this local registry.
-        "artifacts/default".to_string()
+        // The binding name is the routing prefix; the next path segment is the
+        // project namespace (`artifacts/{project_id}` in platform mode,
+        // `artifacts/default` in dev mode). Returning the bare binding name
+        // here — instead of the legacy `"artifacts/default"` — lets the proxy
+        // routing table correctly identify both modes' pushes as local and
+        // extract the right project_id for downstream authz/scope checks.
+        // Matches the per-cloud convention documented in
+        // `internal-docs/alien/02-manager/02-releases.md`.
+        self.binding_name.clone()
     }
 
     async fn create_repository(&self, repo_name: &str) -> Result<RepositoryResponse> {
--- a/crates/alien-cli/src/auth.rs
+++ b/crates/alien-cli/src/auth.rs
@@ -501,8 +501,18 @@ mod oauth_flow {
             project: &'a str,
         }
 
+        // URL-encode the path segment to match the established convention in
+        // the rest of the crate (see `execution_context.rs::resolve_url` and
+        // the gcp / aws clients). Manager IDs are ULIDs today and contain no
+        // URL-special characters, but encoding here keeps the call site
+        // correct-by-construction rather than correct-by-coincidence — a
+        // future change to the ID format won't silently mis-route the request.
         let response = client
-            .post(format!("{}/v1/managers/{}/token", base, manager_id))
+            .post(format!(
+                "{}/v1/managers/{}/token",
+                base,
+                urlencoding::encode(manager_id)
+            ))
             .query(&[("workspace", workspace)])
             .header("Authorization", format!("Bearer {}", user_jwt))
             .json(&ExchangeBody {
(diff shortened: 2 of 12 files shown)


ItamarZand88:
Superseded by #49 — same code, clean history (no force-push trail, no orphan commits with OSS-boundary leaks that earlier rounds introduced).

PR #49, opened 2026-05-24
feat: CLI mints scoped tokens for release; per-project scope on OCI push (ALIEN-139)

ItamarZand88:
## Summary

This PR has two halves that together let `alien release` work safely against a multi-tenant manager. On the CLI side, the user's OAuth token gets exchanged for a short-lived signed token scoped to one project on one manager before any manager call happens — so the manager never sees a credential type it can't validate. On the manager side, the image-push handler now checks that the project scope inside the token actually covers the project being pushed to, so a token issued for project A can't push images intended for project B. OSS standalone behavior is preserved: validators that don't set `Subject.scopes` leave it empty, and the new scope check is a no-op on empty scopes.

Step-by-step flow when you run `alien release`:

1. You ran `alien login` earlier, so the CLI has your OAuth access token saved.
2. The CLI asks the platform for a short-lived signed token scoped to one project on one manager (using the `purpose: registry-push` flow added in the companion platform PR).
3. The CLI sends that signed token to the manager's artifact-registry setup call.
4. The CLI sends the same signed token to the image push at `/v2/.../blobs/uploads/`.
5. **The manager checks the project scope inside the token against the project being pushed to — rejects if the scope doesn't cover this project.** ← the security check this PR adds
6. Push succeeds; the release is created.

This PR changes the OSS side of the release flow: the CLI now asks the platform for a project-scoped signed token instead of sending the raw OAuth token, and the manager's image-push handler now rejects any token whose scope doesn't cover the project being pushed to.

## What was broken before

Three things, all surfaced when running `alien release` against a multi-tenant manager:

1. The CLI was sending the raw OAuth token to the manager. The manager rejected it because OAuth tokens aren't a credential type it can validate.
2. The manager's image-push handler authorized by role only — a token with the right role could push to any project, not just the project it was actually scoped to.
3. `alien commands invoke` errored before reaching the manager at all, because the command code tried to use a client that doesn't exist in platform mode.

## Security review

This PR touches authorization on the image push, so answering the 5 standard questions.

### 1. What's new on the auth path

- **Old:** the push handler ran a role check. A token with the workspace-member role could push to any project in the workspace.
- **New:** if the token carries non-empty project scopes, at least one scope entry must cover the project being pushed to. Tokens with no scopes (the OSS standalone case) skip this check entirely, so single-tenant setups behave exactly as before.

### 2. How the scope check works

In `crates/alien-manager/src/registry_proxy.rs`:

1. The push handler runs the existing role check first (unchanged).
2. If the subject's `scopes` are non-empty, the new check runs: each scope entry is a `workspaceId/projectId` pair, and at least one entry must end in the target project's ID.
3. No scope covers the target → reject with 403.
4. Scopes are empty → skip the new check (OSS standalone behavior preserved).

The scope-check predicate is extracted as a pure function `scopes_cover_project(scopes, project_id)` so it's unit-testable on its own.

### 3. Token shape this PR cares about

The CLI's exchange path produces one token type used for image push:

| What it's for | What it carries |
|---|---|
| Image push during `alien release` | `scopes: ["workspaceId/projectId"]`, manager binding, `purpose: registry-push` |

Token issuing and signature verification happen on the platform side and aren't changed by this PR.

### 4. Where the token is accepted

| Endpoint | Accepts | Extra check this PR adds |
|---|---|---|
| `/v1/artifact-registry/*` | the new exchanged token | none (in this PR) |
| `/v2/.../blobs/uploads/` (image push) | the new exchanged token | the scope must cover the target project's ID |

### 5. What I explicitly checked

- **Project-scope leak:** a token scoped to project A cannot push images to project B. Covered by a test on `scopes_cover_project`.
- **Substring confusion:** a scope ending in `/prj_abc` does NOT accept a target of `prj_abcd` — the suffix match requires the leading `/`. Covered by a dedicated test.
- **OSS standalone preserved:** validators that don't set `Subject.scopes` leave it empty, and the new check is gated on `!scopes.is_empty()`. Verified by a test using an empty-scopes subject, plus a grep — all OSS validator literals initialize `scopes: Vec::new()`.
- **Cache-reuse safety:** the CLI caches the exchanged token; the cache-reuse check (`token_matches_target`) validates the cached token's claims (manager ID, project ID, scope shape) before reuse. 7 tests cover match, manager mismatch, project mismatch, missing claim, malformed token, scopes claim of the wrong type.

### Security follow-up landed on this branch (after re-review)

A follow-up review of the companion platform PR's security work needed the per-project scope predicate from a stable OSS-side location:

- **`949bad8 feat(manager): promote per-project scope check to Subject::permits_project`** — moves the predicate (previously a private free function `scopes_cover_project` in `registry_proxy.rs`) onto `Subject` as a method, so other route handlers in downstream embedders can apply the same per-project invariant without duplicating the logic. Same semantics (empty scopes → `true` for OSS standalone; non-empty → at least one entry ends with `/<project_id>`; substring-confusion guard via the leading slash). The OCI push call site now reads `subject.permits_project(project_id)` instead of `scopes_cover_project(&subject.scopes, project_id)`. Tests move with the function, plus one additional case (multi-scope-any-match).

## Verification

- **How I tested manually:** Ran `alien login` against the local stack, ran `alien init remote-worker-ts my-worker`, applied the template fixes (the template still scaffolds with the old `alien.Function` API instead of `alien.Worker`, and the npm `@alienplatform/core` is stale — separate ticket below), ran `alien link`, then `alien release --experimental` — got two release IDs on two separate runs, the second from a fresh-rebuild scratch state with all local databases wiped. Both passed the new scope check and the artifact-registry + image-push path end-to-end.
- **Unit tests:** `cargo test -p alien-manager` → 108 passing (was 103; the +5 are for the extracted `scopes_cover_project` predicate). `cargo test -p alien-cli --features platform --lib oauth_flow::tests` → 7 tests for `token_matches_target`.
- **Build:** `cargo build -p alien-cli --features platform` → clean.
- **What I couldn't test end-to-end:** `alien commands invoke` against a deployed worker — the C1 fix in this PR unblocks the CLI side, but the full path 401s for a separate reason tracked as **ALIEN-144**.

## Out of scope

- **[ALIEN-144](https://linear.app/alienplatform/issue/ALIEN-144)** — `alien commands invoke` 401 in platform mode. The token this PR's CLI flow produces works for image push, but doesn't work on endpoints that forward the call back to the platform. Needs a different token shape or a server-side exchange. Independent design call.
- `Subject.scopes: Vec<String>` is loosely typed — the `"workspaceId/projectId"` format is only enforced by a doc comment. A typed `permits_project_push(project_id)` helper on `Subject` would replace the open-coded suffix-match at call sites.
- The `remote-worker-ts` template still scaffolds with the old `alien.Function` API and a stale `@alienplatform/core` npm dependency — anyone trying `alien init` followed by `alien release` will hit it. Separate template-update ticket.

## Test plan for reviewer

- [ ] `cargo test -p alien-manager` passes locally (108 tests).
- [ ] `cargo build -p alien-cli --features platform` is clean.
- [ ] Spot-check the 5 new tests on `scopes_cover_project` and the 7 new tests on `token_matches_target` — they cover the security-critical branches.
- [ ] Confirm OSS standalone preserved: `grep -rn 'Subject {' crates/alien-manager/src/providers/` should show `scopes: Vec::new()` on all subject literals.

### DIFF
--- a/crates/alien-bindings/src/providers/artifact_registry/local.rs
+++ b/crates/alien-bindings/src/providers/artifact_registry/local.rs
@@ -147,11 +147,13 @@ impl ArtifactRegistry for LocalArtifactRegistry {
     }
 
     fn upstream_repository_prefix(&self) -> String {
-        // The embedded local registry accepts two-segment repo paths (e.g.,
-        // "namespace/repo"). We use "artifacts/default" as the canonical prefix
-        // — this matches what the CLI hardcodes in dev mode and what the proxy
-        // routing table uses to route pushes to this local registry.
-        "artifacts/default".to_string()
+        // The binding name is the routing prefix; the next path segment is the
+        // project namespace (`artifacts/{project_id}` in platform mode,
+        // `artifacts/default` in dev mode). Returning the bare binding name
+        // here — instead of the legacy `"artifacts/default"` — lets the proxy
+        // routing table correctly identify both modes' pushes as local and
+        // extract the right project_id for downstream authz/scope checks.
+        self.binding_name.clone()
     }
 
     async fn create_repository(&self, repo_name: &str) -> Result<RepositoryResponse> {
--- a/crates/alien-cli/src/auth.rs
+++ b/crates/alien-cli/src/auth.rs
@@ -501,8 +501,18 @@ mod oauth_flow {
             project: &'a str,
         }
 
+        // URL-encode the path segment to match the established convention in
+        // the rest of the crate (see `execution_context.rs::resolve_url` and
+        // the gcp / aws clients). Manager IDs are ULIDs today and contain no
+        // URL-special characters, but encoding here keeps the call site
+        // correct-by-construction rather than correct-by-coincidence — a
+        // future change to the ID format won't silently mis-route the request.
         let response = client
-            .post(format!("{}/v1/managers/{}/token", base, manager_id))
+            .post(format!(
+                "{}/v1/managers/{}/token",
+                base,
+                urlencoding::encode(manager_id)
+            ))
             .query(&[("workspace", workspace)])
             .header("Authorization", format!("Bearer {}", user_jwt))
             .json(&ExchangeBody {
@@ -621,7 +631,16 @@ mod oauth_flow {
             project_id,
         )
         .await?;
-        store_manager_token(&minted)?;
+        // Cache write is best-effort. The mint succeeded, so we hold a valid
+        // token in memory and the caller can complete the operation. A failed
+        // keyring write just means we'll re-mint on the next invocation —
+        // strictly slower, never wrong. This matches the cache-failure pattern
+        // documented in `alien/CLAUDE.md` ("When `warn!` Is Appropriate").
+        if let Err(e) = store_manager_token(&minted) {
+            tracing::warn!(
+                "Failed to cache manager token in keyring (will re-mint next run): {e}"
+            );
+        }
         Ok(minted)
     }
 
--- a/crates/alien-cli/src/auth.rs
+++ b/crates/alien-cli/src/auth.rs
@@ -174,6 +174,10 @@ mod oauth_flow {
     const SERVICE: &str = "alien-cli";
     const ACCESS_USER: &str = "access_token";
     const REFRESH_USER: &str = "refresh_token";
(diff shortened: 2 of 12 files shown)


ItamarZand88:
Closing — restarting from a clean base.

PR #69, opened 2026-06-03
feat: agent self-update across all pull modes — K8s (Helm) + OS-service launcher (Linux/macOS/Windows)

yonatan-alien:
## Summary

Closes ALIEN-59. Manager-driven agent self-update on Kubernetes, plus the two production-hardening fixes that came out of live e2e.

**Feature**
- New `agent_upgrade` loop in `alien-agent` consumes `agent_target.helm` on `/v1/sync` and spawns a Helm-runner Job that runs `helm upgrade --reuse-values --set runtime.image.tag=<target>` under an `alien-agent-upgrader` ServiceAccount.
- `alien-helm` chart guardrails — `required management.{url,name,token,deploymentId}`, `/livez`+`/readyz` probes, Recreate strategy, namespace-scoped upgrader Role/Binding, `upgrader.enabled` opt-out.
- Sync schema gains `agent_version` / `agent_os` / `agent_arch` / `regime` / `agent_image_repository`; K8s worker transport set to `local`; K8s ServiceAccount controller is skipped (chart owns it).
- `alien-manager` exposes `PUT /v1/deployments/:id/target-agent-version`; `build_agent_target` emits `agent_target.helm` when target ≠ reported and regime = kubernetes.

**Production hardening** (would silently break otherwise; both found via live e2e)
- `fix(helm): enable agent runtime-data persistence by default` — without a PVC, the self-update's pod roll wipes the agent's `deployment_id`/sync token, the new pod hits 409 from initialize, CrashLoopBackOffs, and helm `--atomic` rolls back. Documented opt-out.
- `feat(agent,manager): /v1/rejoin endpoint` — for the case where persistence is genuinely off (or the state directory is wiped for any other reason), the agent's initialize-409 falls through to a new `POST /v1/rejoin` that re-acquires a deployment-scoped sync token over the existing row. New `RouterOptions::include_rejoin` + `skip_rejoin()` builder lets multi-tenant embedders override, same pattern as `/v1/initialize`.
- `fix(agent): preserve 409 from initialize so rejoin fall-through fires` — the agent's `.context(ConfigurationError)` was clobbering the SDK's real 409 with a hardcoded 500. New `DeploymentNameAlreadyExists` variant (409) lets the conflict surface so the rejoin branch actually triggers.

Paired changes are tracked in a separate PR.

## Windows OS-service self-update (launcher/operator path)

This branch also lands **Phase 3 (Windows)** of the OS-service self-update path —
the frozen `alien-launcher` supervising `alien-operator` with health-gated binary
swaps + last-stable rollback, now on Windows via the Service Control Manager
(Linux/systemd and macOS/launchd already on the branch). All cross-platform
launcher `core/` logic is shared; only the per-OS shims are new.

- **SCM `ServiceHost` + service entry (T3.1):** `platform/windows.rs` maps SCM
  Stop/Shutdown/Interrogate ↔ the launcher's control channel; `main.rs`'s Windows
  `run_supervisor` hands control to `service_dispatcher` (service mode) or a
  Ctrl-C handler (`--console`, for tests/E2E), then drives the shared `core::run`.
- **Job-Object `ChildSupervisor` (T3.2):** spawns the operator in a kill-on-close
  Job (`command-group`) so it dies with the launcher; graceful stop via
  `CTRL_BREAK`, escalating to Job termination after a grace window.
- **Die-with-parent for the workload + graceful stop (T3.3):** the operator now
  places the user application in its own kill-on-close Job (a new `AppChild`
  wrapper — the Windows counterpart to Linux `PR_SET_PDEATHSIG`) so a self-update
  handoff can't orphan it, and honors `CTRL_BREAK` for a clean shutdown.
- **Junction `VersionStore` (T3.4):** `current`/`last-stable` are directory
  junctions; the non-atomic flip reconciles crash residue on the next start
  (finish-flip / reconstruct-from-probation); `gc` tolerates a locked
  (still-mapped) operator `.exe` by scheduling reboot-deletion instead of failing.
  The full Phase-0 state-machine suite runs unchanged over this store.
- **Installer (T3.5):** `alien-deploy operator install` registers
  `alien-launcher.exe` as the SCM service under `%ProgramData%\alien-operator`
  with a service recovery config (`sc failure`: reset 24h, three 5s restarts);
  junction-seeded layout with exe-suffixed binaries; `--no-launcher` still keeps
  the legacy direct-operator service.
- **Windows CI (T3.7):** a `windows-os-service` job runs the platform unit +
  state-machine tests and a live SCM smoke (`sc create → RUNNING → qfailure →
  stop → delete`, with a stub operator answering `/livez`) on native Windows.

**Deferred (tracked in the internal plan):** Authenticode code-signing — a
cross-platform signing workstream, since nothing is signed on any platform today
— and porting the seven-scenario E2E *rig* to Windows (it's Unix-only: symlinks,
`/bin/sh` version wrappers, `pgrep`/`ps`, a musl workload OCI build). The Windows
platform tests + SCM smoke exercise the same store/child/SCM logic in the meantime.

## Test plan

- [x] `cargo test -p alien-agent -p alien-core -p alien-deployment -p alien-helm -p alien-manager --lib`
- [x] Live e2e against local managerx + orbstack K8s: pin a target version in the dashboard → manager emits `agent_target` → helm-upgrade Job rolls the agent pod → PVC keeps `deployment_id` → new agent reports the new version
- [x] Live e2e for rejoin: `kubectl exec rm -rf /var/lib/alien-agent/*` + bounce pod → agent logs `Name already exists — assuming local state was wiped, rejoining…` → recovers the original `deployment_id` and stays running
- [ ] Reviewer to confirm: the vendored `client-sdks/platform/*` openapi spec + generated SDK — the **v1.13.2 main merge took main's version**, because main's #120/#121 added types (`NewDeploymentRequestInitialDesiredRelease`, `KubernetesClusterSource`, `DeploymentSetupConfigPublicSubdomain`, …) that the auto-merged `alien-cli` source depends on (keeping ours no longer compiled). Consequence: this branch's rejoin / target-agent-version schema additions are currently **dropped from the vendored spec** — the reverse of the earlier merge. The `/v1/rejoin` + target-operator-version manager-side code and the manager SDK are intact; only the vendored SDK mirror is affected. A subsequent **openapi regen must reconcile** so both this branch's additions and main's land in the vendored SDK.
- [x] Windows CI (`windows-os-service` job) green on native Windows: `cargo test -p alien-launcher -p alien-operator` (junction store + Phase-0 state-machine suite + Job-Object kill-on-close + SCM control mapping) **and** a live SCM smoke (`sc create → RUNNING → sc qfailure → sc stop → sc delete`) — run 29160469336


crates/alien-agent/src/loops/agent_upgrade.rs
```diff
@@ -0,0 +1,453 @@
...
+        },
+    });
+
+    (job_name, body)
+}
+
+/// POST a Job to the in-pod Kubernetes API. Returns Ok on 2xx, AlienError
+/// otherwise. 409 (already exists) is treated as success — another sync
+/// already kicked the Job and idempotency handles the retry.
+async fn create_job(namespace: &str, body: &serde_json::Value) -> Result<()> {
```
alongubkin: these APIs should be in alien-k8s-clients

crates/alien-core/src/sync.rs
```diff
@@ -16,6 +16,42 @@ pub struct SyncRequest {
...
     pub current_state: Option<DeploymentState>,
+    /// Agent binary version, from `env!("CARGO_PKG_VERSION")` at build time.
+    /// Lets the manager build fleet-wide version inventory and decide
+    /// whether to send an `agent_target`. Optional for back-compat with
+    /// older agents.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub agent_version: Option<String>,
+    /// `linux` / `macos` / `windows`. From `std::env::consts::OS`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub agent_os: Option<String>,
```
alongubkin: I think we have an existing enum for this somewhere in alien-core

crates/alien-core/src/sync.rs
```diff
@@ -16,6 +16,42 @@ pub struct SyncRequest {
...
+    /// `x86_64` / `aarch64`. From `std::env::consts::ARCH`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub agent_arch: Option<String>,
```
alongubkin: I think we have an existing enum for this somewhere in alien-core

crates/alien-core/src/sync.rs
```diff
@@ -16,6 +16,42 @@ pub struct SyncRequest {
...
+    /// How the agent is supervised — `os-service` (launcher) or `kubernetes`
+    /// (Helm). Detected at runtime from `KUBERNETES_SERVICE_HOST`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub regime: Option<AgentRegime>,
```
alongubkin: regime is a really weird name haha

crates/alien-operator/src/loops/operator_upgrade.rs
```diff
@@ -0,0 +1,453 @@
...
+    } else {
+        format!(" {}", inputs.extra_args.trim())
+    };
+    let helm_cmd = format!(
+        "set -e\n\
+         printf '%s' \"$VALUES_JSON\" > /tmp/values.json\n\
+         exec helm upgrade \"$RELEASE\" \"$CHART_REF\"{version_flag}{extra_args_flag} \
+            --namespace \"$NAMESPACE\" \
+            --reuse-values \
+            --atomic --wait \
```
alongubkin, line 542: Can you document why did we choose `--atomic`?

crates/alien-operator/src/loops/operator_upgrade.rs
```diff
@@ -0,0 +1,453 @@
...
+            --values /tmp/values.json"
+    );
+
+    let body = serde_json::json!({
```
alongubkin, line 546: What happens if creating this job fails? Or pulling the job's image fails? Or the command inside it fails? Do we report back those failures in any way? 

How many retries do we support (I think the answer should be unlimited)? Is there exponential backoff or constant interval between retries?

crates/alien-operator/src/loops/operator_upgrade.rs
```diff
@@ -0,0 +1,453 @@
...
+        "apiVersion": "batch/v1",
+        "kind": "Job",
+        "metadata": {
+            "name": job_name,
+            "namespace": inputs.namespace,
+            "labels": {
+                "app.kubernetes.io/managed-by": "alien-agent",
```
alongubkin: We recently improved our white-labeling support, need to make sure it fits there. See other places in alien-helm

crates/alien-agent/src/loops/agent_upgrade.rs
```diff
@@ -0,0 +1,453 @@
+//! Agent self-update actuator. When `/v1/sync` returns
```
alongubkin: Right now this will only work for Kubernetes deployments, right? No support for local service swap etc? What'll happen in local?

alongubkin (changes requested):

crates/alien-agent/src/loops/agent_upgrade.rs
yonatan-alien: As decided, we'll also include service implementation for Linux/Windows/Mac in this branch before merging.

crates/alien-core/src/sync.rs
yonatan-alien: There is only Architecture for cloud instances (Arm64/X86_64). I'll make enums for the agent arch/os

crates/alien-core/src/sync.rs
yonatan-alien: There is only Architecture for cloud instances (Arm64/X86_64)

crates/alien-core/src/sync.rs
yonatan-alien: I'll change to packaging

PR #78, opened 2026-06-16
fix: allow management role to manage its own inline policies

ab-alien-dev:
The manager re-applies resource-scoped management permissions to the management role via iam:PutRolePolicy when a deployment adds or removes resources after setup. The role previously only had iam:GetRole on itself, so the call failed with:

  CLOUD_PLATFORM_ERROR: Failed to apply management permission
  'worker/dispatch-command' to role '<stack>-management'

Grant the role iam:PutRolePolicy / GetRolePolicy / DeleteRolePolicy / ListRolePolicies, scoped to its own ARN. The role is already the highest-privilege identity in the stack, so this is not a meaningful widening.

Closes ALIEN-198.


alongubkin:
This creates over privileged management roles, closing.

PR #79, opened 2026-06-16
fix(cli): preserve workspace context for platform requests

alongubkin:
## Summary

Fixes ALIEN-199.

Updates the OSS CLI/manager side for the platform workspace selection flow. This complements the platform PR with API enforcement, OAuth user bootstrap, and managerx forwarding.

## Changes

- Use user memberships for CLI workspace discovery so login/bootstrap works before a workspace is selected.
- Keep workspace names user-only in manager subjects instead of adding a top-level field to every subject.
- In API-key mode, do not inherit the saved OAuth default workspace; omit workspace query/header unless `--workspace` was explicitly supplied.
- Keep explicit `--workspace` for API keys as a mismatch check, so a wrong workspace still fails server-side.
- Regenerate Platform OpenAPI copies used by the OSS SDKs from the Platform source schema.

## Generated

- OpenAPI copies were regenerated from Platform source via `pnpm run generate:api-openapi && pnpm run generate:api-copy-openapi` in the Platform repo.

## Validation

- `cargo check -p alien-cli --features platform`
- `cargo check -p alien-cli --no-default-features`
- `cargo test -p alien-cli execution_context --features platform --lib --tests`
- `cargo test -p alien-cli project_link --features platform --lib --tests`
- `cargo check -p alien-managerx`
- `cargo test -p alien-manager subject --lib --tests`
- `cargo test -p alien-managerx token_validator --lib --tests`
- `cargo test -p alien-managerx routes::auth --lib --tests`
- `cargo test -p alien-managerx user_list_deployments_sends_workspace_header_to_platform_api --lib --tests`
- `cargo test -p alien-managerx user_without_workspace_name_fails_before_platform_api_request --lib --tests`
- `cargo test -p alien-managerx set_delete_pending_forwards_caller_token --lib --tests`
- `cargo test -p alien-managerx user_list_releases_sends_workspace_header_to_platform_api --lib --tests`
- `git diff --check`

Live smoke with the matching platform branch via `pnpm dev` against `https://api.alien.localhost`:

- OAuth/user: `alien deployments ls --base-url https://api.alien.localhost --workspace demo --project byoc-database`
- OAuth/user: `alien releases ls --base-url https://api.alien.localhost --workspace demo --project byoc-database`
- OAuth/user default workspace: `alien projects ls --base-url https://api.alien.localhost`
- OAuth/user default workspace: `alien releases ls --base-url https://api.alien.localhost --project alien-private-manager`
- API key, no workspace: local seeded demo key + `projects ls` passed.
- API key, no workspace: local seeded demo key + `--project byoc-database deployments ls` passed.
- API key, no workspace: local seeded demo key + `--project byoc-database releases ls` passed.
- API key, explicit wrong workspace: local seeded demo key + `--workspace alienplatform projects ls` failed with 403 as expected.

## Related

- Requires the matching platform branch/PR for API enforcement, OAuth user auth, and managerx forwarding.


crates/alien-cli/src/commands/platform/workspace.rs
```diff
@@ -124,7 +124,7 @@ pub async fn workspace_task(args: WorkspaceArgs, ctx: ExecutionMode) -> Result<(
 pub async fn list_workspace_names(http: &crate::auth::AuthHttp) -> Result<Vec<String>> {
     let client = http.sdk_client();
     let response = client
-        .list_workspaces()
+        .list_memberships()
```
ab-alien-dev, line 127: Why this change? is this a mistake ?

crates/alien-cli/src/project_link.rs
```diff
@@ -456,20 +456,22 @@ async fn validate_linked_project(
 
 pub async fn get_project_by_name(
     http: &AuthHttp,
-    workspace: &str,
+    workspace: Option<&str>,
```
ab-alien-dev: why is this optional ? isn't a project always on top of a workspace?

crates/alien-cli/src/commands/platform/workspace.rs
alongubkin: Not a mistake. `list_workspaces()` is now workspace-scoped, so it cannot be used during login/workspace bootstrap before the CLI has selected a workspace. `list_memberships()` is the user-scoped bootstrap endpoint for that case; after selection, normal workspace-scoped requests still send `?workspace=...`.

crates/alien-cli/src/project_link.rs
alongubkin: Good catch. A project should always have a workspace. I changed this so `get_project_by_name` takes a required workspace name again; only the outbound `workspace` query is optional. In API-key mode we resolve the workspace name from `whoami`, but omit the workspace query unless `--workspace` was explicitly supplied.

PR #83, opened 2026-06-18
fix(terraform): support network_mode=use-default for EKS, exclude deprecated AZs (ALIEN-202)

ab-alien-dev:
## Summary

Fixes two EKS/Terraform issues that surfaced with `network_mode = "use-default"`, and updates EKS Kubernetes/Helm provider authentication so long applies do not reuse an expired plan-time token.

## Bug 1: Empty `subnet_ids` with `use-default`

`kubernetes_cluster.rs` built the EKS cluster `subnet_ids` ternary with branches for `create-new` and `use-existing` only. `use-default` fell through to `[]`, so the AWS provider rejected apply:

```text
Error: Not enough list items
  on .terraform/modules/deployment/kubernetes.tf line 177
Attribute vpc_config.0.subnet_ids requires 1 item minimum, but config has only 0 declared.
```

`network_mode = "use-default"` was already documented as valid, and the network emitter honored it by skipping subnet creation. The EKS emitter needed matching default-VPC data sources.

## Bug 2: EKS-disallowed Availability Zones

Even after using default-VPC subnets, AWS can return subnets in Availability Zones where the EKS control plane is not supported. AWS documents these restrictions by **Availability Zone ID**, not AZ name, because AZ names are account-local.

The generated Terraform now excludes AWS-documented EKS-disallowed AZ IDs by default:

- `use1-az3`
- `usw1-az2`
- `cac1-az3`

## Fix

For EKS modules, add conditional data blocks gated on:

```hcl
var.kubernetes_cluster_mode == "create" && var.network_mode == "use-default"
```

The generated module now emits:

- `data.aws_vpc.<label>_default { default = true }`
- `data.aws_availability_zones.<label>_eks_supported { exclude_zone_ids = var.unsupported_availability_zone_ids }`
- `data.aws_subnets.<label>_default`, filtered by VPC ID and `availability-zone-id`

`public_subnet_ids_expr` and `private_subnet_ids_expr` route through that filtered default-subnet data source when `network_mode = "use-default"`. Default VPCs only have public subnets, so both expressions intentionally return the same filtered subnet list. Callers that need a real public/private split should use `create-new` or `use-existing`.

The same EKS-safe AZ-ID exclusion is also applied to the `create-new` EKS AZ data source so newly-created EKS subnets do not land in an unsupported AZ.

A new Terraform variable, `unsupported_availability_zone_ids`, defaults to the AWS-documented EKS-disallowed AZ IDs and can be overridden if AWS changes regional support.

## Helm provider auth fix

The EKS Kubernetes/Helm provider config now uses the Kubernetes `exec` auth plugin (`aws eks get-token`) instead of `data.aws_eks_cluster_auth.target.token`.

The token data source resolves at plan time and expires after roughly 15 minutes. A full apply that creates the cluster, provisions IAM, and installs Helm can exceed that window and fail mid-apply with Kubernetes authentication errors. `exec` regenerates a token per Kubernetes API call, matching the generated kubeconfig path.

## Tests

- [x] `cargo check -p alien-terraform`
- [x] `cargo test -p alien-terraform --test generator`
- [x] Added regression coverage for EKS + `NetworkSettings::UseDefault`, including `terraform fmt -check` and `terraform validate`
- [ ] Rebuild packages-builder, bump module version, then run `terraform init -upgrade && terraform apply` against an AWS account whose default VPC includes an EKS-disallowed AZ ID and verify the EKS cluster + Helm install succeed

Closes ALIEN-202.


alongubkin:
I think the `use-default` direction is right, but the AZ exclusion should use **Availability Zone IDs**, not AZ names.

The current implementation defaults `unsupported_availability_zones = ["us-east-1e"]` and feeds it into `exclude_names`. That is account-local: AWS maps AZ names independently for older accounts in several regions, including `us-east-1`, so `us-east-1e` is not a portable identifier for the same physical zone across customers/accounts.

AWS documents the EKS subnet restriction by **Disallowed Availability Zone IDs**, not names:

- `us-east-1` -> `use1-az3`
- `us-west-1` -> `usw1-az2`
- `ca-central-1` -> `cac1-az3`

Docs:
- EKS subnet requirements: https://docs.aws.amazon.com/eks/latest/userguide/network-reqs.html
- AZ IDs and account mapping: https://docs.aws.amazon.com/global-infrastructure/latest/regions/az-ids.html

Terraform has first-class support for the idiomatic version here: `aws_availability_zones` supports `exclude_zone_ids`, and EC2 subnet filtering supports `availability-zone-id`.

Suggested shape:

```hcl
variable "unsupported_availability_zone_ids" {
  type        = list(string)
  description = "Availability Zone IDs to exclude when selecting EKS control-plane subnets."
  default     = ["use1-az3", "usw1-az2", "cac1-az3"]
}

data "aws_availability_zones" "kubernetes_eks_supported" {
  count            = var.kubernetes_cluster_mode == "create" && var.network_mode == "use-default" ? 1 : 0
  state            = "available"
  exclude_zone_ids = var.unsupported_availability_zone_ids
}

data "aws_subnets" "kubernetes_default" {
  count = var.kubernetes_cluster_mode == "create" && var.network_mode == "use-default" ? 1 : 0

  filter {
    name   = "vpc-id"
    values = [data.aws_vpc.kubernetes_default[0].id]
  }

  filter {
    name   = "availability-zone-id"
    values = data.aws_availability_zones.kubernetes_eks_supported[0].zone_ids
  }
}
```

I’d also consider applying the same EKS-safe AZ source to the `create-new` path. Right now `create-new` still creates subnets from `data.aws_availability_zones.available.names[count.index]`, so it can still select an EKS-disallowed AZ ID before the EKS cluster is created.

Finally, please add a generator test that actually renders an EKS module with `network = Some(NetworkSettings::UseDefault)`. The current snapshot updates validate the module syntax, but they do not exercise this new active branch.

alongubkin (approved):

PR #89, opened 2026-06-26
feat(postgres): resource runtime, bindings, and Local controller

ItamarZand88:
The Postgres resource runtime, stacked on the infra fixes (#88): the resource model and bindings, the Local (developer) controller, the cloud client SDKs for the managed Postgres backends, and the TypeScript SDK surface. The part worth careful review is how the generated DB password is handled.

## How the local DB password flows
The password is generated once and has to reach a linked worker for a direct connection, without ever landing in control-plane state. So it runs through two separate channels:

1. **It is stripped from the synced binding params and never written to serialized controller state** — so it cannot reach control-plane storage or status responses.
2. It is handed to the worker at runtime through the worker's environment, resolved per request, never persisted.

Step 1 is the property that matters, and review caught a real gap there: the password was reaching the synced channel. Fixed by stripping it in `get_binding_params` — the `#[serde(skip)]` on the field alone was not enough.

## What's in the layer
- Resource model + bindings (`alien-core`) — the Postgres resource, its binding shapes, the heartbeat data.
- The Local controller (`alien-local` + `alien-infra/src/postgres/local.rs`) — runs an embedded Postgres for local development.
- Cloud client SDKs (`alien-aws-clients` / `gcp` / `azure`) — thin wrappers over the managed cloud Postgres APIs (Aurora, Cloud SQL, Flexible Server).
- SDK surface (`packages/core`, `packages/sdk`) — the generated schemas and the TypeScript binding.

## How I tested
- `cargo test` across the touched crates (`alien-core`, `alien-bindings`, `alien-local`, `alien-infra`), including the binding round-trip and encoding-parity tests.
- The local embedded-Postgres integration test (`alien-local`).
- Exercised end to end in the full Postgres cloud e2e (the setup layer, #90, stacks on this).

Security walk for the password (this PR touches the secret):
- Synced and persisted state never carry the password — the round-trip test asserts it is absent from the serialized binding params.
- The runtime worker-env delivery is per request and not persisted.
- Errors from the secret path are redacted (request body scrubbed before it can reach an error chain).
- The one gap that existed (password on the synced channel) is the one this PR fixes. Nothing else turned up.


ItamarZand88:
Addressed the review feedback in 5a0b31fd (and ran our internal final-review on the changes):

1. **pgvector on boot**: install now skips the networked download when the pinned version is already installed (version-aware, so a `PGVECTOR_VERSION` bump still falls through and reinstalls), so recovering an existing database no longer depends on the release host being reachable.
2. **restart_exited mutex**: collects exited IDs under a brief lock then re-acquires per-database, so the monitor no longer holds `runtimes` across the whole restart batch.
3. **cloud `load_postgres`**: added a trait doc note that cloud Postgres is TypeScript-SDK-only (runtime-rejected, no compile gate). Kept by-design.
4. **`external` ssl**: added a `PostgresConnection.ssl` JSDoc note that it is always `false` for external (node-postgres has no `prefer` mode). Kept as the documented v1 limitation.

alongubkin (approved):

PR #90, opened 2026-06-26
feat(postgres): setup permissions, private networking, and e2e

ItamarZand88:
Postgres setup, stacked on the runtime (#89): the cloud permission sets, the preflights that prepare each cloud, Azure private networking, the secret redaction on cloud create requests, and the end-to-end coverage that exercises Postgres on all three clouds.

What happens when a stack with a Postgres provisions:

1. Preflights enable the required cloud services (Azure / GCP) and provision the secrets vault for the connection password.
2. The permission sets grant the management identity a prefix-scoped set of management-plane actions. **← the heart: management-plane only, never DB-contents access**
3. On Azure, private networking creates the dedicated Private Endpoint subnet (and the importer records it) so the database is reachable only from inside the stack.
4. The end-to-end apps then provision, connect, and run a `pgvector` query against the real cloud database.

This PR adds the setup half of the resource: it turns the runtime model and bindings (#89) into something that can actually be provisioned, secured, and torn down on a real cloud account.

## What's in the layer
- **Permission sets** (`alien-permissions/permission-sets/postgres/*`) — `data-access` (the worker's scoped read of the connection secret), `provision`, `management`, `heartbeat`. The management / provision / heartbeat sets are management-plane only and never grant access to the DB contents.
- **Preflights** (`alien-preflights`) — enable the required cloud services, provision the secrets vault, and assert the network prerequisites before provisioning starts.
- **Azure private networking** (`alien-infra/src/network/azure*`) — the dedicated Private Endpoint subnet and the importer that records it.
- **Cloud create-request redaction** (`alien-client-core/src/request_utils.rs`) — the master password rides the cloud create-request body; it is scrubbed before that body can land in an error chain or synced state.
- **End-to-end coverage** (`alien-test`, `tests/e2e/test-apps/*`) — the comprehensive Rust and TypeScript apps now bind a Postgres and run a query.

## How I tested
- `cargo test` across `alien-permissions` (incl. the AWS ABAC + permission-set validation tests), `alien-preflights`, `alien-infra`, `alien-terraform` (azure generator/snapshots), and `alien-test` — all pass.
- End to end on AWS, GCP, and Azure: provision the database → worker connects → `CREATE EXTENSION vector` + a `pgvector` query succeeds → teardown. Run from the comprehensive Rust and TypeScript test apps.

Security walk (this PR touches permissions and the secret):
- The management-plane permission sets grant no DB-contents access — the sensitive-data validation tests assert this.
- The master password on a cloud create request is redacted before it can reach a persisted error or synced state.
- Permission scopes are pinned to the stack / resource prefix, never broad cloud admin.
- Nothing turned up.


ItamarZand88:
Addressed the review feedback in 7b8b829e (and ran our internal review on the changes):

1. **AWS `resource` binding subgrp ARN**: added `subgrp:${stackPrefix}-${resourceName}-*` to the resource binding, matching the per-resource subnet group name (`{prefix}-{id}-subnets`); the resource-scoped role can now manage its subnet group lifecycle.
2. **Imported network None PE subnet**: this is not silent. The Postgres controller reads it with `.ok_or_else` and fails fast with an actionable error (set privateEndpointSubnetName, or let Alien manage the VNet). Clarified the importer comment; carrying the configured PE subnet through import is a follow-up.
3. **Azure provision missing `read`**: documented the intent. The `/read` omission is deliberate (the controller confirms each PE/DNS-zone op via the Azure-AsyncOperation LRO header, not a GET), kept minimal per least-privilege.

alongubkin (approved):

PR #132, opened 2026-07-19
ci: harden CI against recurring flakes (apt, GCP quota, sccache, Node 20)

lilienblum:
## Why

Dispatching **Cloud Tests** against current main fails on environment noise rather than product code. Four distinct, demonstrated failure classes:

1. **GCP per-minute quota exhaustion** — [main run 29666176524](https://github.com/alienplatform/alien/actions/runs/29666176524): `test_comprehensive_iam_policy_operations` died creating its test service account with HTTP 429 `RESOURCE_EXHAUSTED` ("Service accounts created per minute per project", `RetryInfo { retryDelay: 60s }`). The test project is shared with concurrently running CI jobs (e.g. E2E), so any run can lose this race. The fail-fast cancellation then hid 20 more tests.
2. **Ubuntu mirror outages during apt installs** — [run 29437605942](https://github.com/alienplatform/alien/actions/runs/29437605942): the `bindings` job spent 9 minutes in apt connection errors installing `protobuf-compiler` and died with exit 100 before running a single test. The same un-retried apt pattern exists in every workflow. The apt pattern has burned Cloud Tests twice (also May 19, [run 26096349837](https://github.com/alienplatform/alien/actions/runs/26096349837)).
3. **sccache cache-backend errors** — the shared sccache webdav backend intermittently answers a read with a transient `403` (`PermissionDenied (temporary)`), which aborts sccache server startup and fails the whole Rust build with exit 101. By default sccache treats a server it cannot reach as fatal.
4. **Node-20 action deprecation** — every `actions/*` pinned to a Node-20 runtime is currently being force-migrated to Node 24 by GitHub and will eventually hard-fail once Node 20 is removed from the runners.

## What

- **Bounded quota retry at the test service-account creation sites** (`tests/common/mod.rs`, used by the resource-manager and IAM suites, including the duplicate-creation test so throttling can't masquerade as the expected 409). Retries only `RateLimitExceeded`, max 3 attempts, waiting out Google's documented 60s RetryInfo window. This is server-side throttling on a shared project — not a client or test bug — so honoring the server's retry contract with bounded attempts keeps real failures visible.
- **Install protoc from GitHub releases instead of Ubuntu mirrors** via `arduino/setup-protoc@v3` pinned to `27.x` — the action + version the Windows E2E job already uses — across ci-fast, ci-fork, cloud-tests, generated-drift, e2e-cloud, release, and the `setup-rust` composite action.
- **`Acquire::Retries=5` on the remaining apt installs** (musl-tools, awscli/unzip, release-pgvector build tools).
- **`SCCACHE_IGNORE_SERVER_IO_ERROR=1`** at the workflow level (ci-fast, cloud-tests, generated-drift, release, e2e-cloud) so sccache falls back to local compilation when the shared cache backend is unreachable instead of failing the build. ([sccache README](https://github.com/mozilla/sccache/blob/main/README.md))
- **De-deprecate Node-20 actions** to their current Node-24 majors: `checkout@v5`, `setup-node@v5`, `cache@v5`, `download-artifact@v7` (deliberately not the breaking `v8`, which stops auto-unzipping and errors on hash mismatch), `upload-artifact@v6`, `pnpm/action-setup@v6`, `setup-terraform@v4`, `setup-helm@v5`, `paths-filter@v4`, `sccache-action@v0.0.10`. Each target tag was verified to declare `runs.using: node24`.

## Validation

- `cargo test -p alien-gcp-clients --features gcp --no-run` compiles all touched test binaries.
- CI Fast on this PR exercises the new protoc install and the bumped actions on the ARM runners.
- Every bumped action tag verified against its `action.yml` to declare the `node24` runtime; `download-artifact` held at `v7` after reviewing the `v8` breaking changes.
- Will dispatch **Cloud Tests** on this branch and label the PR `run-e2e` to exercise the e2e-cloud changes end-to-end.

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -40,14 +40,20 @@ runs:
       run: |
         git config --global url."https://x-access-token:${{ inputs.repo-access-token }}@github.com/".insteadOf "https://github.com/"
 
-    - name: Install musl-tools and protoc
+    - name: Install protoc
+      if: inputs.install-protoc == 'true'
+      uses: arduino/setup-protoc@v3
+      with:
+        version: "27.x"
+        repo-token: ${{ github.token }}
+
+    - name: Install musl-tools
       shell: bash
+      # Retry transient Ubuntu mirror failures; a dead mirror has killed CI
+      # jobs before (apt exit code 100 after minutes of connection errors).
       run: |
-        PACKAGES="musl-tools"
-        if [ "${{ inputs.install-protoc }}" = "true" ]; then
-          PACKAGES="$PACKAGES protobuf-compiler"
-        fi
-        sudo apt-get update && sudo apt-get install -y $PACKAGES
+        sudo apt-get -o Acquire::Retries=5 update
+        sudo apt-get -o Acquire::Retries=5 install -y musl-tools
 
     - name: Add Rust targets
       if: inputs.targets != ''
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -28,7 +28,7 @@ runs:
 
     - name: Install sccache
       continue-on-error: true
-      uses: mozilla-actions/sccache-action@v0.0.9
+      uses: mozilla-actions/sccache-action@v0.0.10
       with:
         # Avoid the action's latest-release API lookup, which can fail during
         # transient GitHub API outages. Cache setup is an optimization, so a
--- a/.github/workflows/ci-fast.yml
+++ b/.github/workflows/ci-fast.yml
@@ -90,7 +90,10 @@ jobs:
           bun-version: "1.3.14"
 
       - name: Install protoc
-        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+        uses: arduino/setup-protoc@v3
+        with:
+          version: "27.x"
+          repo-token: ${{ github.token }}
 
       - name: Install CloudFormation linter
         run: pipx install cfn-lint
--- a/.github/workflows/ci-fast.yml
+++ b/.github/workflows/ci-fast.yml
@@ -24,8 +24,10 @@ jobs:
       # reports "runner lost communication" with no logs). Full-width
       # parallelism (16 rustc jobs, then 16 concurrent test binaries — many
       # of which spawn servers and embedded databases) is what spikes it.
-      CARGO_BUILD_JOBS: "12"
-      NEXTEST_TEST_THREADS: "10"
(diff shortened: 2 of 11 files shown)


lilienblum:
Validation complete — all three runs green on this branch:

- **CI Fast** ([29667071841](https://github.com/alienplatform/alien/actions/runs/29667071841)): full build + unit tests pass with protoc 27.x from `arduino/setup-protoc` on the ARM runners.
- **Cloud Tests** ([29667075765](https://github.com/alienplatform/alien/actions/runs/29667075765)): all 5 jobs green, including the `gcp-clients` job that failed on main in [29666176524](https://github.com/alienplatform/alien/actions/runs/29666176524). This run didn't hit quota contention, so the retry stayed dormant; it only engages on a real 429.
- **E2E Tests** ([29667077516](https://github.com/alienplatform/alien/actions/runs/29667077516)): AWS push + local Linux agent flows green with the reworked protoc/apt steps.

For reference, a baseline E2E dispatch on main ([29666177279](https://github.com/alienplatform/alien/actions/runs/29666177279)) also passed, so Cloud Tests' quota flake was the only genuine main-branch failure this change needed to address.

lilienblum:
Added f19fac04: **lower CI Fast parallelism (`CARGO_BUILD_JOBS` 12→10, `NEXTEST_TEST_THREADS` 10→8)**.

While validating this PR, the latest main push ([run 29707403037](https://github.com/alienplatform/alien/actions/runs/29707403037), #144) failed CI Fast **twice in a row** — both attempts hit the documented memory-starvation mode (runner lost communication mid Rust-tests, no logs, no failing step recorded). Two consecutive identical deaths on the same commit means the test set has outgrown the 12/10 bounds chosen when that mitigation was added, so this steps both down one notch. Main itself stays red until this merges.

lilienblum:
Correction: the parallelism reduction (f19fac04) is reverted in c67838e5 — **the runner-lost failures were a GitHub Actions platform incident, not memory starvation.**

githubstatus.com currently shows an unresolved **"Incident with GitHub Actions — Partial Outage"** (plus degraded API requests) that started Jul 19; every "runner lost communication" death (both attempts of main run [29707403037](https://github.com/alienplatform/alien/actions/runs/29707403037) at 23:07/23:56 UTC and this branch's run [29709108728](https://github.com/alienplatform/alien/actions/runs/29709108728) at 00:11 UTC) falls inside that window, right after three consecutive green ~30-minute main runs. Slowing CI permanently based on outage artifacts would be wrong, hence the revert — the PR is back to net quota-retry + protoc/apt hardening only.

I'll rerun this PR's checks and main's failed run once the incident resolves.

PR #138, opened 2026-07-19
feat: add email resource backed by AWS SES

alongubkin:
## Summary

Adds `alien.Email` — email infrastructure for sending and receiving mail on customer-owned domains, backed by AWS SES.

### Design: infrastructure vs bindings vs runtime data

Email-sending products typically manage domains at runtime: create an SES identity via API when a customer adds a domain, hand the DNS records to the customer, poll verification, delete the identity when the domain is removed. The resource is designed around that split:

- **CloudFormation-managed infrastructure**: the shared `AWS::SES::ConfigurationSet`, the event topology (config set → SNS topic → linked SQS queue), the inbound receipt topology (`AWS::SES::ReceiptRuleSet` + a **catch-all** `AWS::SES::ReceiptRule` writing into the linked Storage bucket), and one `AWS::SES::EmailIdentity` (Easy DKIM) per **optional seed domain**. `domains` defaults to empty — a config-set-only resource is valid; seed domains exist for day-0 bootstrap and static-domain products.
- **Capability grants**: `email/send` (send through any identity + the stack-scoped config set), `email/manage-identities` (the runtime domain-management capability: create/get/list/delete identities, Put MailFrom/DKIM/feedback/config-set attributes), `email/provision` (deploy-time), `email/management` (read-only).
- **Minimal binding**: `{ service: "ses", region, configurationSet, eventTopicArn? }`. Deliberately no domain list — identities come and go at runtime, so a deploy-frozen list would be stale by design; applications discover the current set via `ses:ListEmailIdentities`.
- **Runtime identities are application data**: identities created through the manage-identities grant are not tracked by CloudFormation and are not removed on stack deletion; the application owns their lifecycle, including deletion. Documented on the resource rustdoc, the emitter, and the TS builder.

The inbound receipt rule omits `Recipients` entirely: SES treats a recipient-less rule as matching every recipient the account receives mail for, so mail addressed to runtime-verified identities lands in the bucket with zero infrastructure mutation.

### What it provisions (AWS)

- One shared `AWS::SES::ConfigurationSet` (named `${AWS::StackName}-<id>`); per seed domain, an `AWS::SES::EmailIdentity` with Easy DKIM enabled and the config set associated.
- With `.events(queue)`: an `AWS::SES::ConfigurationSetEventDestination` (SEND / DELIVERY / BOUNCE / COMPLAINT / DELIVERY_DELAY / REJECT) → `AWS::SNS::Topic` → `AWS::SNS::Subscription` to the linked SQS queue (raw delivery), plus the `AWS::SQS::QueuePolicy` that lets the topic send.
- With `.inbound(storage)`: an `AWS::SES::ReceiptRuleSet` + catch-all `AWS::SES::ReceiptRule` whose `S3Action` writes raw incoming mail into the linked Storage bucket. The `ses.amazonaws.com` write grant (conditioned on `aws:SourceAccount`) is emitted as part of the storage bucket policy — S3 supports only one bucket-policy resource per bucket, so the storage emitter owns it. The receipt rule `DependsOn` the bucket policy because SES validates write access at rule-creation time.

### Builder surface (TypeScript)

```ts
const mailbox = new alien.Storage("mailbox").build()
const mailEvents = new alien.Queue("mail-events").build()

// Domain-dynamic products: no seed domains; identities are managed at
// runtime through the email/manage-identities grant.
const email = new alien.Email("mailer").inbound(mailbox).events(mailEvents).build()

// Static-domain products can seed identities at deploy time:
const seeded = new alien.Email("mailer").domains(["mail.example.com"]).build()
```

`domains` is append-friendly: adding a seed domain provisions a new identity, removing one deletes its identity (and DKIM verification state); the list may be empty. `inbound`/`events` can be added, removed, or repointed.

### DNS records in outputs

Domain verification is asynchronous and owned outside the deployment, so outputs carry the records instead of a verification result: `EmailOutputs` maps each seed domain to its three Easy-DKIM CNAME tokens (`{ name, value }` from `DkimDNSTokenName1..3` / `DkimDNSTokenValue1..3`), plus `configurationSet` and (when inbound is configured) `ruleSetName`. The import ref wires these via `Fn::GetAtt`. For runtime-created identities the application gets the DKIM tokens from the `CreateEmailIdentity` / `GetEmailIdentity` responses.

### ⚠️ Receipt rule set activation caveat

SES allows only **one active receipt rule set per AWS account**, and CloudFormation has no resource that activates one (verified against current CFN docs — the only workarounds are custom-resource lambdas, deliberately not done in v1). Activation is a documented post-deploy step:

```
aws ses set-active-receipt-rule-set --rule-set-name <ruleSetName>
```

This is documented on the resource rustdoc, the TS builder, the emitter, and surfaced via the `ruleSetName` output. SES email receiving is also region-limited; deploying `inbound` to an unsupported region fails at the CloudFormation layer (documented, not hard-failed).

### Permissions

`email/` permission sets (AWS-only in v1):

- `provision` — SES identities/config sets/receipt rules + SNS/SQS wiring + S3 bucket policy (deploy-time).
- `send` — `ses:SendEmail`/`SendRawEmail` on `identity/*` (runtime identities must be sendable) + the stack-prefixed configuration-set ARN.
- `manage-identities` — **the runtime domain-management capability**: `CreateEmailIdentity`, `GetEmailIdentity`, `DeleteEmailIdentity`, `PutEmailIdentityMailFromAttributes`, `PutEmailIdentityDkimAttributes`, `PutEmailIdentityFeedbackAttributes`, `PutEmailIdentityConfigurationSetAttributes` on `identity/*`, plus `ListEmailIdentities` on `*` (no resource-level support per the Service Authorization Reference — it is the discovery path replacing the binding's domain list).
- `management` — read-only.

ABAC-validation exceptions are documented in the validation tests: SES identities are named after customer mail domains (unknowable at deploy time, no stack prefix possible), and the SES v1 receipt-rule actions do not support resource-level permissions per the Service Authorization Reference.

Ownership: `email` is **Frozen-only** (setup owns the durable routing state end to end).

### Scope notes

- Runtime send bindings are intentionally out of scope for v1 — workers send with the AWS SDK directly using the `email/send` grant.
- Manager-side import (`AwsEmailImportData` + `alien-infra` importer/controller) is a follow-up; importing a stack that contains an email resource will report `ImportRegistrationMissing` until then.
- If two email resources point `events` at the same queue, their queue policies would overwrite each other; use one events queue per email resource (documented in the emitter).

## Validation

- Unit tests: builder (including the no-seed-domains default), `validate_update` semantics (id immutable; adding/removing/emptying domains and links allowed), dependency extraction, serde round-trips (empty `domains` omitted from serialization).
- Generator tests: dedicated email scenarios (full wiring with catch-all inbound assertion, minimal, config-set-only, import/binding refs, live-linked-queue rejection) plus the full-stack audit template includes Email with inbound + events. Every rendered template passes `cfn-lint` (no errors; only pre-existing informational W3005 dependency warnings on the full-stack template).
- **Real AWS deploys** (test account, us-east-1, throwaway domains):
  - Initial deploy of the full rendered template — all 12 resources reached `CREATE_COMPLETE`; verified DKIM tokens, config-set association, S3 action, and SES's `AMAZON_SES_SETUP_NOTIFICATION` bucket-policy probe. Stack deleted and cleanup confirmed.
  - Re-deployed after the catch-all change to confirm SES accepts a **recipient-less receipt rule via CloudFormation**: `CREATE_COMPLETE`, `describe-receipt-rule` shows no `Recipients` key (catch-all semantics), SES again wrote its S3 probe object, and the account's pre-existing active receipt rule set was untouched. Stack deleted, retained bucket removed, and identities/config set/rule set confirmed gone.
- `cargo fmt`, clippy on touched crates, `cargo test` for alien-core / alien-cloudformation / alien-permissions (all green), `@alienplatform/core` build + typecheck + tests, `biome check`.
- Regenerated TS schemas via `pnpm generate` in `packages/core` (`domains` is now optional in the generated Zod/JSON schema; the earlier run also picked up a pre-existing drift in `AzureComputeClusterImportData.identityId`).

Note: `alien-terraform::generator azure_network_create_emits_vnet_subnets_nat` currently fails on a stale snapshot (Azure container-apps subnet) on `main` too — unrelated to this change, left untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


alongubkin:
## Overnight live-deploy additions (all on this branch)

Three more commits landed after live-deploying a stack that uses this resource through the platform:

1. **Importer** (`AwsEmailImporter`): maps the CloudFormation `emit_import_ref` payload into an `AwsEmailController` pinned at `Ready`; covered by importer round-trip tests (wire-shaped payloads, typed `EmailOutputs`, missing-field error cases).
2. **Manager-delivered binding** (`get_binding_params`): workers linking an imported email resource booted without `ALIEN_<ID>_BINDING` and crashed — the controller now delivers `{service: ses, configurationSet, region}` (region recorded at import; optional-on-read for states imported before the field existed). Live-verified: the linked Lambda workers boot and read the binding.
3. **Resource-scoped permission emission** (`email_iam_policies`): grants of `email/send` / `email/manage-identities` on the resource were silently dropped by the emitter — senders failed with `not authorized to perform ses:SendRawEmail`. Policies now attach to the owning service-account roles; template-parsing test asserts actions, statements, and role attachment.

Also: the compile-time preflight allowlist gained `email` (earlier commit), and the rendered template was validated end-to-end through a real CloudFormation deploy + platform import.

PR #139, opened 2026-07-19
feat: experimental AWS OpenSearch Serverless resource

alongubkin:
Adds `alien.experimental.AwsOpenSearch` — the first resource in the new `experimental/` namespace — provisioning **next-generation Amazon OpenSearch Serverless** (collection groups, scale-to-zero) via CloudFormation.

## Experimental namespace convention

- Resource type identifiers under `experimental/` (this resource: `experimental/aws-opensearch`) mark provider-specific resources that don't abstract over clouds and may change in breaking ways before being promoted to portable resources. Documented in `crates/alien-core/src/resources/mod.rs` and the resource rustdoc.
- `ResourceType` is a plain string, permission-set ids already use `/` (`storage/data-read`), and the derived ids (`experimental/aws-opensearch/provision`) flow through the existing `format!("{type}/provision")` machinery unchanged, so the `/` separator needed no parser changes.
- AWS-only enforcement: only an AWS CloudFormation emitter is registered. Any other platform fails with the typed `IMPORT_REGISTRATION_MISSING` error at generation time (covered by a test).
- In the TypeScript SDK the resource lives under a new `experimental` sub-namespace: `new alien.experimental.AwsOpenSearch("articles").collectionType("vectorSearch")`.

## Next-gen ground truth (verified live in us-east-1, 2026-07-19)

- **CLI**: `aws opensearchserverless` exposes the full next-gen surface (`create-collection-group`, `batch-get-collection-group`, `create-collection --collection-group-name`, control-plane `create-index`/`get-index`).
- **CloudFormation**: the public registry contains `AWS::OpenSearchServerless::CollectionGroup` (with `Generation: CLASSIC|NEXTGEN`), plus `CollectionGroupName`/`EncryptionConfig` on `Collection`, and `CollectionIndex`/`Index` types (all last updated 2026-07-16). So collection groups **are provisionable via CloudFormation today** → the emitter provisions next-gen directly (decision rule (a)).
- Live findings baked into the emitter:
  - `StandbyReplicas: DISABLED` is **rejected** for `NEXTGEN` groups ("StandbyReplicas cannot be set to DISABLED for NEXTGEN collection groups"), so the group pins `ENABLED` and the knob is not exposed (next-gen has replication built in; scale-to-zero makes the classic toggle moot).
  - Next-gen collections have **no `DashboardEndpoint`**: `Fn::GetAtt` fails the deploy and `batch-get-collection` omits the field, so outputs expose only `endpoint` and `collectionArn`.
  - No separate encryption `SecurityPolicy` is needed for next-gen: `EncryptionConfig: { AWSOwnedKey: true }` on the collection suffices.

## What gets provisioned (AWS emitter)

- `AWS::OpenSearchServerless::CollectionGroup` — `Generation: NEXTGEN`, `StandbyReplicas: ENABLED`, Retain deletion policy (durable search state; the group must outlive its retained collection).
- `AWS::OpenSearchServerless::Collection` — `SEARCH` or `VECTORSEARCH`, joined to the group, AWS-owned-key encryption, Retain.
- Network `SecurityPolicy` — public endpoint; access is still SigV4 + IAM + data-access-policy gated.
- Data `AccessPolicy` + `aoss:APIAccessAll` IAM policies — emitted for every service-account role whose permission profile grants `experimental/aws-opensearch/data-access` on this resource, with the IAM statement pinned to the collection ARN via `Fn::GetAtt` (AOSS collection ARNs use server-assigned ids, so the static permission set can only scope to `collection/*`). Skipped when no role holds a grant.
- Physical names are `{id}-{8-char stack suffix}`; the id is validated at generation time against the AOSS name grammar (lowercase, 3–23 chars) so failures are clear errors instead of mid-deploy rollbacks.

## Config surface

- `id` (start_fn) and `collectionType: search | vectorSearch` (default `search`, immutable — `validate_update` rejects changes). That's intentionally the whole surface: capacity knobs don't apply to scale-to-zero next-gen, and `TIMESERIES` is not exposed.
- Ownership: frozen-only (setup-owned durable state) in `ownership_policy_for_resource_type`.
- Permission sets: `experimental/aws-opensearch/provision` (control plane, derived from the CFN handler permissions) and `experimental/aws-opensearch/data-access` (`aoss:APIAccessAll`; fine-grained index/document rights are enforced by the data-access policy).
- Binding payload: `{ service: "aoss", endpoint, collectionName }` — `aoss` is the SigV4 signing service name (not `es`). Body-carrying requests additionally need `x-amz-content-sha256` (verified live: AOSS returns an empty 403 without it). Runtime data bindings (SDK client wrappers) are intentionally out of scope — workers use SigV4 HTTP with the granted IAM.

## Validation

- Unit tests: builder defaults, resource-type constant, `validate_update` immutability (id, collectionType), serde round-trips (Rust + TS).
- Generator tests: full-template snapshot (group + collection + policies + IAM wiring), vector-search type, id-grammar rejection, missing-emitter error on GCP/Azure/Kubernetes. `cfn-lint` (1.53.0, which knows the new registry types) passes on every render.
- **Live AWS validation** (account 230470760195, us-east-1, all resources deleted afterwards):
  1. Deployed the exact emitted template → `CREATE_COMPLETE`; collection reached `ACTIVE` during stack creation inside a `generation: NEXTGEN` group with `minIndexing/SearchCapacityInOCU: 0` (scale-to-zero confirmed via `batch-get-collection-group`).
  2. Data plane proven with the exact emitted data-access-policy JSON: control-plane `create-index`, SigV4 `POST /_doc` → `201 created`, `POST /_search` → `200` with the document returned (first query `took` ≈ 11 s — the next-gen cold start; subsequent ≈ 200 ms).
  3. Created/deleted: 2 CFN stacks, 2 collections, 2 collection groups, 2 network policies, 2 data-access policies, 1 test index. Verified zero remaining collections/groups/security-policies/access-policies.
- `cargo nextest` green across the touched crates (alien-core / alien-cloudformation / alien-permissions / alien-preflights / alien-terraform / alien-helm, 906 tests) and the rest of the workspace except the suites CI also excludes (live cloud-client tests requiring `.env.test`); fmt + clippy clean; TS build + typecheck + vitest + biome green.

Generated artifacts (`packages/core/src/generated/**`) regenerated with `pnpm generate` in `packages/core`; regeneration also picked up a previously uncommitted `identityId` field on `AzureComputeClusterImportData` from main's Rust source.

## Incidental fixes (needed to keep CI green)

- **ABAC validation**: documented exceptions in `aws_abac_validation.rs` for the AOSS resource model — control-plane actions can't be resource-scoped (policies have no ARNs, collection ids are server-assigned), and the `data-access` grant's `collection/*` bound is narrowed to the exact ARN by the emitter. `iam:CreateServiceLinkedRole` is a separate statement conditioned on `iam:AWSServiceName: observability.aoss.amazonaws.com`.
- **iam-dataset pin bump** (`permission_set_validation.rs`): the pinned commit predates the next-gen `aoss:*CollectionGroup*` actions; bumped to a current commit that includes them (all 118 permission tests pass on the new pin).
- **Pre-existing red test on main**: `azure_network_create_emits_vnet_subnets_nat` fails on `origin/main` (verified in a clean worktree) because #135 added the Container Apps subnet without updating the snapshot. Snapshot updated here in a separate commit so this PR's CI can pass.

## Follow-ups

- Stack-import/heartbeat controller in `alien-infra` (the resource is frozen-only; importer + heartbeat land with the controller work).
- Optional `AWS::OpenSearchServerless::Index` support for declaring indexes (next-gen made indexes a control-plane object).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


alongubkin:
## Overnight live-deploy additions (all on this branch)

Two more commits after live-deploying a stack that uses this resource through the platform:

1. **Importer** (`AwsOpenSearchImporter`): maps `emit_import_ref` output into an `AwsOpenSearchController` pinned at `Ready` (collection name/id/ARN/endpoint); importer round-trip tests with wire-shaped payloads and typed outputs.
2. **Manager-delivered binding** (`get_binding_params`): workers linking the imported collection booted without their binding — the controller now delivers `{service: aoss, endpoint, collectionName}`, mirroring the emitter's binding ref.

Live validation on a real deployment: SigV4 (`aoss`) search requests from the linked workers succeed against the NEXTGEN serverless collection (first request after idle pays the scale-to-zero resume, ~30s, then sub-second). The compile-time preflight allowlist gained `experimental/aws-opensearch` in an earlier commit.

PR #154, opened 2026-07-20
feat: grant remote storage data access

lilienblum:
Summary

- Grant the management identity storage/data-write only for Storage resources with remoteAccess.
- Keep every grant scoped to the concrete resource.
- Apply the behavior consistently to Auto, Extend, and Override management profiles.
- Leave non-remote storage unchanged.

Verification

- cargo fmt --check
- cargo test -p alien-preflights management_permission_profile --lib
- git diff --check

This is an independently reviewable foundation for remote customer-cloud storage bindings.


lilienblum:
Consolidated into #156 so the repository has one review surface for the complete remote storage bindings change. This commit is preserved there as its own commit.

PR #155, opened 2026-07-20
fix: gate imported remote binding params

lilienblum:
Summary

- Publish imported binding parameters only when the resource explicitly enables remoteAccess.
- Skip binding extraction entirely for non-remote imports.
- Align imported-state behavior with the normal provisioning executor.
- Cover both remote and non-remote importer behavior.

Verification

- cargo nextest run --all-features -p alien-infra --test importers
- cargo fmt --check
- git diff --check


lilienblum:
Consolidated into #156 so the repository has one review surface for the complete remote storage bindings change. This commit is preserved there as its own commit.

PR #178, opened 2026-07-22
feat(ai): add the ai resource with an OpenAI-compatible gateway

ItamarZand88:
## Summary

Adds an `ai` resource that gives a workload an OpenAI-compatible endpoint for model inference, in two modes. By default it's keyless: for the host cloud's own models — Bedrock on AWS, Vertex on GCP, Foundry on Azure — a loopback gateway signs each request with the workload's ambient cloud identity, so there are no model API keys to store. A workload can instead bring its own provider key (e.g. OpenAI), in which case the client talks to that provider directly with the vault-resolved key and the gateway isn't involved.

What happens on the keyless path, when a workload declares an `ai` binding for one of the host cloud's models:

1. A loopback AI gateway starts next to the app and its URL is placed in the environment; the app points a normal OpenAI-compatible client at it.
2. **The gateway translates each OpenAI-style request into the target cloud's native model call — Bedrock `InvokeModel`, Vertex `rawPredict`, or the Foundry Anthropic endpoint — and signs it with the workload's ambient cloud credentials.** ← the heart of the change
3. The response, including streamed tokens, is translated back to the OpenAI/Anthropic shape and returned to the app.

For a cloud's own models, this changes AI access from managing model API keys to using the cloud account the workload already runs in.

## What I did

- Add a gateway engine (Rust) that translates OpenAI-shaped requests to each cloud's model API and signs them with ambient credentials.
- Add the `ai` resource to the SDK, plus the controllers and per-cloud setup that grant a workload access to that cloud's models.
- Support a bring-your-own-key mode: the same `ai` binding can point at an external provider with the workload's own key (vault-resolved, redacted in logs), which bypasses the gateway and calls the provider directly.
- Scope the ambient model-invocation grant to inference only: no deployment or data-plane writes.
- Ship the gateway as one standalone binary everywhere: containers run it as their entrypoint, the SDK spawns it on first use and reads back the URL it prints, and compiled single-file Workers embed the binary and extract and spawn it at runtime.
- Make `getAvailableModels` report real availability: the gateway probes each catalog model with a one-token request under the workload's own inference credential and lists only what the cloud actually has enabled (a 429 counts as enabled but rate-limited; 400/401/403/404 drop the model). Each entry carries `provider` and `displayName` so an app can build a model picker, and the example README documents each model's one-time enablement step per cloud.
- Type the client surface: `chat.completions.create` and `responses.create` return OpenAI's own result types (a types-only dev dependency), so callers read `.choices[0].message.content` with no cast.
- Add an `ai-quickstart-ts` example built around "list the available models, then pick one".

## Files touched

- `crates/alien-ai-gateway` — the gateway engine (router, model catalog, availability probe, per-cloud translation + signing) and the `alien-ai-gateway` launcher binary.
- `packages/ai-gateway` — the TypeScript wrapper: resolves and spawns the gateway binary, and the ambient vs BYO-key connection resolution.
- `crates/alien-core`, `packages/core`, `packages/sdk` — the `ai` resource type, the model catalog with per-model activation metadata, and the SDK surface (plus embedding the gateway binary into compiled Workers via `packages/package-layout` and `crates/alien-build`).
- `crates/alien-infra` — the AI controllers and runtime wiring.
- `crates/alien-terraform` + `crates/alien-cloudformation` — the per-cloud setup emitters and their snapshots.
- `crates/alien-permissions/permission-sets/ai/*` — the `ai/invoke` grant (plus heartbeat/management/provision).
- `examples/ai-quickstart-ts` — a runnable example.

## How I tested

- **Manually:** From a deployed test worker, real Claude requests went through the gateway with no API keys configured — Bedrock (AWS) and the Foundry Anthropic endpoint (Azure) both returned completions, including streaming. I also ran the gateway on my machine against real Bedrock and hit `/v1/models`: 36 of the 42 AWS catalog models came back listed and enriched, and the 6 dropped were Claude ids that account/region cannot serve.
- **Compiled Worker:** `bun build --compile` of a Worker that calls `ai()` — the embedded gateway binary is extracted and spawned from inside the single-file binary and serves a loopback URL (the package-layout compile smoke).
- **Unit tests:** cargo nextest across the workspace (CI-fast filter), including the availability-filter suite (mock upstream: an enabled model is kept and enriched, a 403 model is dropped, a 500 keeps the model and re-probes on the next call, a second call is served from cache); SDK vitest + biome; `terraform validate` on the AI setup-emitter output and the CloudFormation snapshot suite; generated-schema drift check clean. The live Bedrock test reads the shared test env and runs in the credentialed job.
- **E2E (ran green on real Bedrock):** the comprehensive TypeScript app lists every model `getAvailableModels` returns and invokes each one, failing if a listed model is not invocable. That run passed after it caught a real filter bug: Bedrock answers `400 "The provided model identifier is invalid"` for a model the account cannot address, and I had been counting 400 as available, so it now drops the model (400 joins 401/403/404) and the list only ever contains invocable models.
- **Anything I couldn't test:** Vertex (GCP) live inference is pending a model-garden entitlement on the test project — unrelated to this code; the GCP emitter path is covered by snapshots + `terraform validate`.

### One non-AI fix rides along

The cloud e2e for this PR overflowed its test thread's stack during deploy. It is a pre-existing issue, independent of the AI code: the deployment reconcile polls a deep async chain (executor to controllers to the cloud SDKs' tower/hyper/rustls futures) that needs about 3MB in a debug build, over tokio's 2MB default thread stack. Release binaries fit 2MB and are unaffected. `RUST_MIN_STACK=8MB` for dev/test threads via `.cargo/config.toml` fixes it, matching the headroom `alien-worker-runtime` already gives its runtime. It rides along here because this PR's own e2e could not run without it.

- **Over-scoped model grant.** `ai/invoke` grants inference only — AWS `bedrock:InvokeModel*` (+ mantle `CreateInference` scoped to `project/default`, not `project/*`); a GCP custom role of just `aiplatform.endpoints.predict`/`explain` (not `roles/aiplatform.user`, which carries deploy + dataset-export); Azure "Cognitive Services OpenAI User" (data-plane only). No deployment writes or data reads. `crates/alien-permissions/permission-sets/ai/invoke.jsonc`.
- **Availability probe surface.** The `/v1/models` probe sends a one-token inference request signed with that same inference-only grant; it adds no control-plane permission, cannot enable or subscribe anything, and its logs carry only the model id and status, never request or response bodies. `crates/alien-ai-gateway/src/availability.rs`.
- **Credential exposure.** On the keyless path the gateway signs with the workload's ambient cloud identity — no key is stored. On the BYO-key path the provider key is a vault-resolved secret: its `Debug` impl redacts it (`crates/alien-core/src/bindings/ai.rs`) and it's resolved into the worker environment, not logged or held in control-plane state. `crates/alien-ai-gateway/src/config.rs`.
- **Off-host reachability.** The gateway binds the loopback interface and hands the app its URL out of band, so it isn't reachable from other hosts. `crates/alien-ai-gateway/src/lib.rs`.
- **BYO-key providers.** An external (bring-your-own-key) provider is explicitly not served by the ambient-credential gateway path. `crates/alien-ai-gateway/src/config.rs`.

Nothing turned up.


examples/ai-quickstart-ts/alien.ts
```diff
@@ -0,0 +1,26 @@
+import * as alien from "@alienplatform/core"
+
+// A model-less AI resource. The customer's cloud serves the inference; the
```
alongubkin: "model-less AI resource" - confusing
'embedded gateway workload ambient identity etc -- implementation details remove from example

examples/ai-quickstart-ts/alien.ts
```diff
@@ -0,0 +1,26 @@
...
+// embedded gateway injects the workload's ambient identity, so no API keys.
+const llm = new alien.AI("llm").build()
+
+const api = new alien.Worker("api")
+  .code({ type: "source", src: "./", toolchain: { type: "typescript" } })
+  .publicEndpoint("api")
+  // Linking injects ALIEN_LLM_BINDING and exposes the gateway as ALIEN_AI_GATEWAY_URL.
```
alongubkin: no need for this comment to be here

examples/ai-quickstart-ts/src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  return c.json({ models: models.map(m => m.id) })
+})
+
+// One-shot question -> answer. `?model=` overrides the default (first catalog entry).
+app.get("/ask", async c => {
+  const question = c.req.query("q")
+  if (!question) {
+    return c.json({ error: "pass a question as ?q=..." }, 400)
+  }
+  const llm = ai("llm")
```
alongubkin: `ai("llm")` is very confusing :) lets think of better name than "llm"

examples/ai-quickstart-ts/src/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  const model = c.req.query("model") ?? (await llm.getAvailableModels())[0]?.id
+  if (!model) {
+    return c.json({ error: "no models available for this cloud" }, 500)
+  }
+  const completion = (await llm.chat.completions.create({
+    model,
+    messages: [{ role: "user", content: question }],
+  })) as { choices?: Array<{ message?: { content?: string } }> }
```
alongubkin: this code is a bit messy.. why is this necessary? need a much cleaner API

crates/alien-ai-gateway/src/lib.rs
```diff
@@ -0,0 +1,178 @@
+//! Embedded, protocol-agnostic AI gateway: a loopback HTTP server that injects the
```
alongubkin, line 1: Let's rename this crate to `alien-ai-gateway`

crates/alien-gateway/tests/live_bedrock.rs
```diff
@@ -0,0 +1,148 @@
+//! Live verification against real AWS Bedrock. Ignored by default — it makes a
+//! real inference call and needs ambient AWS credentials in the environment.
```
alongubkin: Let's not ignore these tests by default, they're very important (same for foundry vertex etc). And lets run them and make sure they're all passing.

You can load our `alien-test-target` account credentials from ../env.test - see this pattern in the alien-bindings tests.

crates/alien-gateway/tests/live_bedrock.rs
```diff
@@ -0,0 +1,148 @@
...
+//!
+//! Run it with:
+//!   eval "$(aws configure export-credentials --profile <p> --format env)"
```
alongubkin: no need for this thing once we'll load ../env.test

crates/alien-ai-gateway-node/src/lib.rs
```diff
@@ -0,0 +1,52 @@
+//! Node-API addon for the alien AI gateway.
```
alongubkin: Hmmm.. let's think if we really need this crate. Maybe we can just start an `./alien-ai-gateway` process from the typescript package?

discuss a little bit with claude what are the advantages / disadvantages of this

packages/ai-gateway/PACKAGE_LAYOUT.md
```diff
@@ -0,0 +1,77 @@
+# `@alienplatform/ai-gateway` — package layout contract
```
alongubkin: what is all this? needed?

packages/ai-gateway/package.json
```diff
@@ -0,0 +1,57 @@
+{
```
alongubkin, line 1: Let's think about it for a second. Do we really need a separate package for this? or should this just be inside @alienplatform/bindings?

alongubkin (changes requested):

packages/sdk/tsdown.config.ts
```diff
@@ -14,14 +14,21 @@ export default defineConfig({
...
+  // embedded-addon registration used by compiled binaries). Inlining a copy here
+  // would give the SDK its own loader instance, so a compiled worker's
+  // `installEmbeddedAddon()` (which registers into the real modules via their
+  // `/native` subpaths) would not be seen by the SDK's re-exported
+  // `kv`/`storage`/`queue`/`vault` (bindings) or `ai`/`getAiConnection`
+  // (ai-gateway). Keeping them external makes the compiled binary share one
+  // module for each. They're real runtime dependencies, so the package stays
+  // self-contained.
+  external: [
+    "@alienplatform/bindings",
```
alongubkin, line 28: do we really need both @alienplatform/bindings and @alienplatform/ai-gateway and not put them on the same thing? is there very large impact on the binary size?

alongubkin (changes requested):

packages/ai-gateway/package.json
ItamarZand88: Looked at folding this into `@alienplatform/bindings` and decided to keep it separate.

Main reason: bindings is in every worker (kv/storage/queue/vault), so if the gateway lives there too, every worker ships the `alien-ai-gateway` binary even when it never calls `ai()`. Keeping it on its own means only AI workers carry it.

They're also different beasts now. bindings is a napi addon loaded in-process, the gateway is a standalone binary we spawn. Different build and staging (it's why alien-build has the Napi/Binary split). One package for both felt off.

Re binary size: same conclusion. There's only one napi addon left now (bindings), and the gateway binary only gets embedded into compiled workers that actually use `ai()`, so a normal worker doesn't grow. Merging is what would bloat everything.

Left it separate for now, lmk if you disagree.

packages/sdk/tsdown.config.ts
ItamarZand88: Kept them separate (more on the package.json thread). The size question actually argues for it: the gateway binary only lands in compiled workers that call `ai()`, so merging into bindings would put it in every worker instead. No size win from combining.

packages/ai-gateway/src/client.ts
```diff
@@ -0,0 +1,453 @@
...
+    params: ResponseCreateParams & { stream: true },
+  ): Promise<AsyncIterable<ResponseStreamEvent>>
+}
+
+// Upstream base URL (no `/v1`) for a BYO-key provider. `ALIEN_AI_LOCAL_BASE_URL` overrides it so
+// any OpenAI-compatible provider works; unknown providers default to OpenAI's.
+function providerBaseUrl(provider: string): string {
+  const override = process.env.ALIEN_AI_LOCAL_BASE_URL
+  if (override) return override.replace(/\/$/, "")
+  return provider === "anthropic" ? "https://api.anthropic.com" : "https://api.openai.com"
```
lilienblum: **🟠 High.** An unvalidated `provider` can silently send a BYO key to the wrong host.

`provider` is a free-form string at every layer (`ExternalAiBinding.provider: String`, `binding.ts` `z.string()`, no allowlist), and `providerBaseUrl` here falls back to `https://api.openai.com` for anything that isn't exactly `"anthropic"`. `_postSurface` then attaches the BYO key as `Authorization: Bearer` to that base. So a binding with `provider: "Anthropic"` (casing), `"google"`, `"groq"`, or a typo, and no `ALIEN_AI_LOCAL_BASE_URL` override, ships the customer's key to OpenAI. Only the `openai` BYO path is tested, so these paths are unexercised.

**Suggested fix.** Validate `provider` against a known set at the resolve boundary and fail closed instead of defaulting to OpenAI.

_(Correcting my earlier note: the `"anthropic"` branch itself is fine. `https://api.anthropic.com/v1/chat/completions` with Bearer is Anthropic's OpenAI-compat endpoint. The one real gap on that path is `responses.create`, since the compat layer doesn't serve `/v1/responses`, so a BYO-anthropic Responses call would 404 while chat/completions works.)_

packages/ai-gateway/src/gateway.ts
```diff
@@ -0,0 +1,223 @@
...
+function keepChildAlive(child: ChildProcess, onGatewayLost: () => void): void {
+  const onGone = () => {
+    if (liveChild === child) liveChild = undefined
+    onGatewayLost()
+  }
+  child.on("error", onGone)
+  child.on("exit", onGone)
+  child.stdout?.resume()
+  child.stderr?.resume()
+  child.unref()
```
lilienblum, line 238: **🟠 High (scoped to one-shot processes).** `child.unref()` here does not let a short-lived host process exit.

`unref()` drops only the process handle, but the `resume()` calls just above put stdout/stderr into flowing mode, which keeps those pipe handles referenced. I reproduced this with a detached long-running child: after `resume()` + `child.unref()`, the parent was still alive until the child exited. Long-running Workers want to stay up anyway, but a CLI, batch job, or test process that calls `ai()` can hang forever.

**Suggested fix.** Unref the stdout/stderr pipe handles after draining them, or restructure stdio so no referenced pipes remain after the readiness handshake. Note that `ChildProcess.stdout`/`stderr` are typed as `Readable`, so direct `child.stdout?.unref()` does not type-check even though the runtime pipe is a `net.Socket`; use a type-safe socket/narrowing helper. Add a one-shot subprocess test that starts the gateway and asserts the host exits.

crates/alien-ai-gateway/src/router/mod.rs
```diff
@@ -0,0 +1,2137 @@
...
+    let routes: HashMap<String, GatewayRoute> =
+        routes.into_iter().map(|r| (r.name.clone(), r)).collect();
+    let models_cache =
+        routes.keys().map(|name| (name.clone(), tokio::sync::OnceCell::new())).collect();
+    let state = Arc::new(AppState {
+        routes,
+        client: reqwest::Client::new(),
+        models_cache,
+    });
+    Router::new()
```
lilienblum, line 87: **🟡 Should-fix (easy).** Not merge-blocking, but it silently rejects valid large requests.

The handlers extract `body: Bytes` and the router never calls `DefaultBodyLimit::disable()`, so axum 0.8's default 2 MB limit 413s requests before they reach the upstream. That's reachable for vision/base64 images and long tool-heavy conversations.

**Suggested fix.** Since this is a pure proxy, disable the limit and let the upstream enforce its own: `Router::new().layer(DefaultBodyLimit::disable())`.

crates/alien-terraform/src/emitters/gcp/ai.rs
```diff
@@ -0,0 +1,39 @@
+//! GCP AI — Vertex AI inference gateway.
+//!
+//! Vertex AI is a project-level service with no per-stack resource to
+//! provision. The emitter returns an empty fragment and carries the project
+//! and location in the import ref so the controller can reconstruct the
+//! Vertex AI endpoint without a cloud round-trip.
+//!
+//! The `aiplatform.googleapis.com` API enablement is handled by the
+//! `GcpServiceActivationEmitter` when the preflight injects a
```
lilienblum, line 9: **🟡 Should-fix (easy).** Breaks the Frozen/Terraform GCP path only (the native controller is fine).

The doc here says aiplatform enablement is handled by `GcpServiceActivationEmitter` via an injected `ServiceActivation`, but `get_required_services()` in `gcp_service_activation.rs` has no `"ai"` arm, so nothing injects it. `terraform apply` still succeeds (the emitter just produces no `google_project_service`); the failure surfaces later at gateway runtime as `SERVICE_DISABLED` when Vertex is first called on a project that didn't already have the API on. The native controller enables it at runtime (`ai/gcp.rs`), so only Terraform-mode GCP deploys are affected.

**Suggested fix.** Add an `"ai"` arm to `get_required_services()`.

lilienblum (commented):
Did a deeper pass on the non-example parts. Two high-severity issues (a BYO provider-validation gap, and a process-lifecycle bug that hangs one-shot callers) plus two cheap fixes (a 2 MB proxy body limit and a Frozen-vs-Live GCP gap). The gateway proxy core itself looks solid.

packages/ai-gateway/src/client.ts
```diff
@@ -0,0 +1,453 @@
...
+}
+
+// A small curated chat-model list for the BYO-key picker (we don't proxy the provider's own
+// /v1/models, which returns hundreds of non-chat entries).
+function defaultModels(provider: string): string[] {
+  return provider === "anthropic"
```
lilienblum, line 111: **🟠 High (BYO Anthropic).** `getAvailableModels()` returns only two retired model families for this provider.

`claude-3-5-sonnet-latest` and `claude-3-5-haiku-latest` point at Claude 3.5 families whose pinned models were retired on 2025-10-28 and 2026-02-19 respectively. Anthropic documents that requests to retired models fail: https://platform.claude.com/docs/en/about-claude/model-deprecations. Because `staticModels` bypasses the provider models endpoint, a BYO-Anthropic picker receives only unusable IDs and the first real completion fails.

**Suggested fix.** Return active IDs such as `claude-sonnet-5` / `claude-haiku-4-5` and cover the Anthropic branch with a request test. Better, query the provider models endpoint so this list does not silently age out again.

crates/alien-ai-gateway/src/creds.rs
```diff
@@ -0,0 +1,515 @@
...
+            }),
+            // `managed` only ever builds a Gcp / Azure native source.
+            BearerSource::Static { .. } | BearerSource::Managed { .. } => None,
+        })
+    }
+
+    /// Cache-then-fetch the workload's projected-identity token from the instance metadata
+    /// service. `source` must be `Gcp` or `Azure`.
+    async fn metadata_token(&self, source: &BearerSource) -> Result<String> {
+        if let Some((tok, exp)) = self.cache.lock().await.as_ref() {
```
lilienblum: **🟠 High on projected-identity GCP/Azure workloads.** This cache miss is not single-flight.

The mutex protects only the lookup and is released before the metadata request. The first `/v1/models` call uses `join_all` to launch every model probe concurrently (14 on GCP, 12 on Azure); each probe authorizes independently, so they can all observe an empty cache and hit IMDS at once. The same stampede happens when a token expires under concurrent inference. If IMDS throttles, model probes become indeterminate, the list is deliberately left uncached, and the next call repeats the fan-out.

**Suggested fix.** Serialize refreshes behind the cache lock or store a shared in-flight refresh future/`OnceCell`, then add a concurrent cold-cache test that asserts exactly one metadata request.

crates/alien-ai-gateway/src/router/mod.rs
```diff
@@ -0,0 +1,2137 @@
+//! The pure proxy: route a request to the model's native cloud endpoint, inject the
```
lilienblum, line 1: **🟡 Structural.** This new module lands at 2,137 lines and has too many reasons to change.

The first ~974 lines combine shared routing/transport, cloud URL selection, three provider adapters, Bedrock request normalization, and an event-stream decoder; another ~1,160 lines of tests live in the same file. `availability.rs` already imports several of these internals, so the boundary is starting coupled rather than cohesive.

**Suggested fix.** Decompose before this becomes the permanent extension point: keep dispatch/common transport in `router/mod.rs`, move Bedrock, Vertex, and Foundry handling into focused modules, and isolate the Bedrock event-stream codec with its tests. That leaves future provider changes local and keeps each module reviewable.

PR #179, opened 2026-07-23
fix(aws): harden CloudFormation BYOC releases

alongubkin:
## Summary

- emit valid NextGen OpenSearch Serverless and storage CORS CloudFormation resources
- preserve deployer input values and include transitive native-addon bytes in build cache keys
- avoid unauthenticated bootstrap OTLP export when runtime secrets are vault-backed
- parse raw storage keys before signing so reserved characters are encoded exactly once

## Validation

- focused OpenSearch and storage CloudFormation generator tests
- AWS CloudFormation ValidateTemplate and real create/update deployment proof
- real AWS NextGen collection-group deployment: `Generation: NEXTGEN`, zero minimum indexing/search OCU, `on.aws` endpoint
- full locally built application proof through Platform `pnpm dev`: TLS/health, send, asynchronous indexing, full-text search hit, and content retrieval
- focused native storage-path and transitive-cache tests
- focused worker-runtime OTLP test plus deployed cold-start/export proof
- cargo check --locked -p alien-bindings-node

Linear: ALIEN-320


alongubkin:
Final AWS validation update:

- Live OpenSearch Serverless rejected `1` OCU for both indexing and search; the supported nonzero floor is `2`. Validation/schema and the CloudFormation fixture now enforce that before deployment (`4229ed78`, `3a45a908`).
- The generated CloudFormation setup path updated a real NEXTGEN collection group to min indexing/search `2.0` OCU; the stack reached `UPDATE_COMPLETE`. The runtime manager correctly refused to mutate the frozen setup-owned capacity.
- Real Lambda testing exposed that an 8-second pre-poll readiness budget could still cross Lambda’s 10-second init limit. `22c1bcc3` now polls the Runtime API immediately and keeps readiness waiting inside invocation handling. A cold API invocation had one init (`1334.03 ms`), HTTP 200, and no init timeout/restart.
- A cold events invocation initialized once (`1165.78 ms`) and a deliberately unmatched valid SQS event returned `FunctionError: Unhandled` / `TASK_DELIVERY_FAILED`, proving failures are no longer silently acknowledged.
- Focused results: OpenSearch core 8/8, CloudFormation generator 7/7, TypeScript core 67/67, AWS runtime 80/80. Fresh CI is running on `3a45a908`.

alongubkin:
@greptileai

alongubkin:
@greptileai

alongubkin:
The remaining 4/5 finding is fixed in 35400f50: `valid_nonzero_ocu` no longer accepts 1, and the invalid-range test now explicitly covers `maxOcu: 1`. This matches the schema and the live AWS validation result. @greptileai

alongubkin:
@greptileai The generated API schema is now synchronized from the source generator, removing the erroneous nullable component wrapper that broke the Rust SDK and CLI.

alongubkin:
@greptileai review

PR #180, opened 2026-07-23
chore: stop building packages during install

lilienblum:
## Summary

- remove the `prepare` build hook from the core, commands, bindings, and SDK packages
- keep package compilation behind the existing explicit `build` commands

## Why

A workspace `pnpm install` currently compiles all four packages as a lifecycle side effect. Bootstrap only needs dependencies installed, so that work is unnecessary and makes setup appear stuck.

The stable and dev npm release workflows already run `pnpm build` before publishing or packing. A `prepack` replacement would keep compilation implicit and duplicate those release build steps.

## Validation

- fresh and repeated `pnpm install --frozen-lockfile` complete without invoking the package builds
- explicit builds succeed for core, commands, bindings, and SDK
- package-layout fixture passes all 28 packed-artifact and consumer checks
- Biome passes for the changed manifests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -520,6 +520,11 @@ jobs:
       - name: Install JavaScript dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Build packages used by addon smoke test
+        run: |
+          pnpm --filter @alienplatform/core build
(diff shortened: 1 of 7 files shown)


lilienblum:
Superseded by #236, recreated from the same tested head on an ALIEN-374-compliant branch so CI can run.

PR #199, opened 2026-07-24
feat: gate resources in the setup generators, not the emitters

ItamarZand88:
## Summary

Converting a resource type to `.enabled()` used to mean hand-threading the gate through every one of its emitters — a full PR per type, four times so far. The setup generators now apply gating as a post-pass over plain emitted blocks, and every existing gated snapshot renders byte-identically to prove the swap changed nothing.

What happens when a gated stack renders:

1. One policy in `alien-core` decides whether the gate is allowed — checked by the preflight and re-checked inside both generators, so rendering without preflights refuses the same things.
2. The generator puts the gate on every owned block (a Terraform `count`, a CloudFormation `Condition`), rewrites every reference to a gated address into indexed form — expressions and string templates alike — and fails generation if any gated address survives unindexed in the rendered files. **← the heart**
3. The resource's registration entry splices out behind the input, while shared support blocks (project-wide roles) stay ungated by classification and body-identical grants keep merging their contributors' gates.

This PR changes gating from per-emitter hand-written code to one generator-owned pass.

## What was broken

Nothing was misbehaving — the cost was structural. Gating knowledge lived inside each emitter (gate calls, `[0]` indexing on self-references, an opt-in flag), so every new resource type re-paid the same conversion work, and a half-converted emitter was only caught by its own flag.

## What I did

- Added `alien_core::gateability`: a type table plus an id-level rule (the reserved `secrets` vault), enforced at preflight and at render in both generators, and exported as a generated manifest that a TypeScript test holds the SDK's `.enabled()` surface against. This PR pins it to today's gateable set — nothing new becomes gateable here.
- Built the Terraform post-pass (`gating.rs`): gates owned blocks, rewrites references (traversals, string templates, `%{if}`/`%{for}` directives) module-wide, refuses `count`/`for_each` collisions, and scans the rendered output as a last net. Splats and `depends_on` stay unindexed — they are already list-aware.
- Classified instead of gated: fragments mark shared blocks (GCP custom roles, Azure role definitions) that must outlive any one resource; footprintless blocks (`random_id`) come from an exact central allowlist and are refused if they contain a `provisioner`; Email's SES statement inside the storage bucket policy became a contribution carrying Email's gate (renders unchanged while Email is ungated).
- Deleted `supports_enabled_when()` from both emitter traits and stripped all gate threading from the kv, queue, and storage emitters across AWS, GCP, and Azure in both formats.
- Added a matrix test on both formats: every registered emitter must be policy-refused or covered by a gated fixture that renders, validates, and resolves both answers. Its first run caught 15 emitter cells that only the deleted opt-in flag had been holding back; the policy now names them.
- Vault became the first type gated purely by the post-pass — no vault emitter ever carried gating code — locked by per-cloud snapshots.

## Files touched

- `crates/alien-core/src/gateability.rs` (new), `lib.rs`, `bin/schema_exporter.rs` — the policy and the generated manifest
- `crates/alien-preflights/src/compile_time/resource_enabled_valid.rs`, `src/mutations/secrets_vault.rs` — ban arms replaced by the shared policy; the vault-id constant moved to core
- `crates/alien-terraform/src/gating.rs` (new), `generator.rs`, `emitter.rs`, `emitters/enabled.rs`, `registry.rs`, emitters under `aws/`, `gcp/`, `azure/` — the post-pass, fragment classification, registry iteration, threading removal, input-name collision check
- `crates/alien-cloudformation/src/generator.rs`, `emitter.rs`, `emitters/aws/{kv,queue,storage}.rs`, `registry.rs` — policy check replaces the opt-in; the SES contribution; parameter collision check
- Tests: `gating_matrix_tests.rs` in both crates (new), gated-vault snapshots (new), collision and policy-refusal tests
- `packages/core/src/__tests__/gateability.test.ts` (new), `src/generated/gateability.json` (new)

## How I tested

- **Byte-parity:** every pre-existing gated snapshot (kv/queue/storage across AWS/GCP/Azure Terraform and AWS CloudFormation) renders identically under the post-pass — `cargo test -p alien-terraform -p alien-cloudformation` with zero snapshot diffs.
- **Matrices:** the registry walk renders every allowed (type × platform × format) gated from a fixture, runs `terraform validate` / cfn-lint plus the deploy-time condition-resolution model, and asserts a declined resource leaves no owned blocks, no grants binding it, and no registration entry.
- **Suites:** alien-core 440, alien-preflights, alien-terraform 120, alien-cloudformation 65 (the two `aws_open_search` cfn-lint failures also fail on untouched main — a stale local lint spec), SDK vitest 71 plus a clean `tsc`.

- A declined resource leaving a reachable grant behind — every formerly hand-gated IAM block (role policies, bucket policies, IAM members, role assignments) is owned and gated by the post-pass; the unchanged per-block gating tests and the matrix assert it.
- Declining one resource deleting a shared role its siblings still hold — shared blocks are never gated, and same-address different-body shared blocks fail generation instead of one silently winning.
- A caller skipping preflights gating a refused type — the generators re-check the policy; a render-side test proves the refusal.
- A reference escaping the rewrite and silently mis-resolving when the deployer declines — the rendered-output scan fails generation with a non-retryable error; unit-tested including splat, key-index, and directive edge cases.
Nothing turned up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,7 +85,7 @@ dependencies = [
 
 [[package]]
 name = "alien-aws-clients"
-version = "2.0.2"
+version = "2.1.6"
 dependencies = [
  "alien-aws-clients",
  "alien-client-core",
@@ -127,7 +127,7 @@ dependencies = [
 
 [[package]]
 name = "alien-azure-clients"
-version = "2.0.2"
+version = "2.1.6"
 dependencies = [
  "alien-client-core",
  "alien-core",
@@ -171,7 +171,7 @@ dependencies = [
 
 [[package]]
 name = "alien-bindings"
-version = "2.0.2"
+version = "2.1.6"
 dependencies = [
  "alien-aws-clients",
  "alien-azure-clients",
@@ -215,7 +215,7 @@ dependencies = [
 
 [[package]]
 name = "alien-bindings-node"
-version = "2.0.2"
+version = "2.1.6"
 dependencies = [
  "alien-bindings",
  "alien-error",
@@ -232,7 +232,7 @@ dependencies = [
 
 [[package]]
 name = "alien-build"
-version = "2.0.2"
+version = "2.1.6"
 dependencies = [
  "alien-build",
  "alien-core",
@@ -266,7 +266,7 @@ dependencies = [
 
 [[package]]
 name = "alien-cli"
-version = "2.0.2"
+version = "2.1.6"
 dependencies = [
  "alien-bindings",
  "alien-build",
@@ -337,7 +337,7 @@ dependencies = [
 
 [[package]]
 name = "alien-cli-common"
-version = "2.0.2"
+version = "2.1.6"
 dependencies = [
  "alien-core",
  "alien-deployment",
@@ -351,7 +351,7 @@ dependencies = [
 
 [[package]]
 name = "alien-client-config"
-version = "2.0.2"
+version = "2.1.6"
 dependencies = [
  "alien-aws-clients",
  "alien-azure-clients",
@@ -369,7 +369,7 @@ dependencies = [
 
 [[package]]
 name = "alien-client-core"
-version = "2.0.2"
+version = "2.1.6"
 dependencies = [
  "alien-error",
  "anyhow",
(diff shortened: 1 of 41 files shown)


ItamarZand88:
Superseded by the same commits on a branch named per the PR validator; reopened as a new PR.

PR #202, opened 2026-07-24
feat: gate every user resource — compute, email, and search

ItamarZand88:
## Summary

`.enabled()` now covers every user resource: compute gates as a live existence choice, email and search gate at setup — each admitted with zero type-specific gating code, which is what the generator-owned mechanism exists for. This PR also pins the semantics: a frozen gate is answered once and enforced for the deployment's lifetime, and a declined live workload keeps its derived baseline so acceptance can return.

What happens when a deployer changes a gate's value on an update:

1. A value conflicting with a recorded frozen answer refuses the update before anything runs (`FROZEN_GATE_ANSWER_CHANGED`); a re-registration that flips one fails synchronously, so a CloudFormation parameter edit rolls the whole stack update back.
2. Frozen declines strip before the mutations — a declined setup resource never existed, nothing derives from it — while live declines strip after them, so service accounts, profile grants, and capacity stay identical whether a live workload is on or off. **← the heart**
3. The executor's ordinary create/delete planning applies the answer: declining deletes the resource (data included), accepting recreates it, exactly like removing or re-adding it from a release.

This PR changes the gateable set from four store types to every user resource, with frozen answers fixed and live answers reversible.

## What was broken

- The two deployment paths disagreed on ordering: the initial path derived infrastructure before stripping declines, the update path after — so declining a live workload on an update would have rederived its service account away and tripped the frozen-compatibility check.
- A frozen gate on a direct deployment (no setup template) ignored the input entirely and created the resource regardless of the deployer's answer.
- Nothing recorded a frozen gate's answer, so nothing could refuse a later change — it was silently ignored rather than an error.

## What I did

- Widened the gateability policy: worker, daemon, and container gate live; email and the experimental search collection gate frozen. The matrices grew fixtures for each, a live-only arm proving a gated live worker is invisible to setup while the deployer is still asked for the input, and a proof that the SES inbound grant follows Email's gate through both deploy-time answers. Builders for all five expose `.enabled()`, held to the policy by the generated manifest.
- Split the decline strip around the mutations on both paths, resolving frozen declines from the import when one exists (absence is the answer) and from the initial input values on direct deployments — never a guess.
- Recorded the canonical frozen answers on the deployment's runtime metadata at creation (runner and import paths both), refused conflicting updates and re-registrations, and derived answers from settled state for deployments that predate the recording. Deliberately no protocol version bump: the field is additive, an older actor's strip already resolves frozen presence from state so it cannot flip a resource, and a bump would refuse every customer-scheduled pull agent the moment a newer manager writes state.
- A gated worker consuming an ungated queue passes preflight — pausing the consumer is the point, and the queue's retention policy owns the backlog. The reverse direction stays refused.
- Live-gate transitions log a structured audit event with an operation id derived from resource, input, answer, and release — identical across retries, fresh for a later flip, no stored counter.
- Extended the enabled-demo distribution app with a gated worker pair on dedicated permission profiles: the accepted worker's function must exist, the declined worker's function must never be created, and both profiles' service accounts must exist — the persisted baseline.

## Files touched

- `crates/alien-core/src/gateability.rs`, `deployment/state.rs` — policy widening; `GateAnswers` on the runtime metadata
- `crates/alien-deployment/src/{pending,updating,lib,error}.rs` — the split strip, answer resolution and enforcement, audit events, typed refusals
- `crates/alien-manager/src/routes/stack.rs`, `tests/stack_import.rs` — import-time recording and the synchronous re-import refusal
- `crates/alien-preflights/src/compile_time/resource_enabled_valid.rs` — compute ban removed; pause-consumer tests both directions
- `packages/core/src/{worker,daemon,container,email}.ts`, `experimental/aws-opensearch.ts`, tests, manifest — the SDK surface
- `crates/alien-deployment/tests/test_platform.rs` — worker decline/re-accept baseline test; frozen-answer lifetime test
- `tests/e2e/test-apps/enabled-demo/*`, `crates/alien-test/{src,tests}/distribution.rs` — the gated worker pair and its cloud assertions

## How I tested

- **State machine:** `cargo test -p alien-deployment --test test_platform` drives the full cycles — a live worker declined and re-accepted across updates with its service account asserted present in the prepared stack throughout, and a frozen answer recorded, refused on conflict, accepted on match, and still refused after wiping the recorded answers to simulate an older deployment.
- **Import:** `cargo test -p alien-manager --test stack_import` — a fresh import records the declined and accepted answers; a re-registration that flips one gets a synchronous 400 with `FROZEN_GATE_ANSWER_CHANGED`.
- **Suites:** alien-core, alien-preflights, alien-terraform, alien-cloudformation, alien-deployment, alien-manager all green; SDK vitest 71 plus clean `tsc`.
- **Real cloud:** three local runs of `terraform_aws_push_enabled_demo` each failed during image upload with an ngrok gateway error (`Service Unavailable`, before Terraform ran — no gating code in the path, no cloud resources created), so the run is dispatched on this branch in the cloud e2e workflow with the `enabled-demo` app filter; the PR should merge on that run's green.

- A changed input creating or deleting a frozen resource — two independent layers: the explicit refusal, and the strip resolving from state presence rather than the new value; the lifetime test proves both directions.
- An answer flip riding an authorized setup update — the re-registration check runs before setup-update authorization is minted.
- A declined live workload leaving reachable access — its resource-scoped grants leave with it; what persists is the provisioning baseline (service account, provision grants) that re-acceptance requires, which the e2e asserts explicitly.
- Mixed-version actors — an older actor ignores a flipped value (state-derived strip, no resource churn) instead of applying it; answers dropped by an old write-back are rebuilt from state presence.
Nothing turned up.

## Out of scope

- Container decline coverage through the local harness, and scale-to-zero for capacity groups whose residents are all declined — capacity keeps its declared floor here.
- Frozen postgres gating — postgres has no setup emitters yet; its gate stays live-only until it does.


ItamarZand88:
Real-cloud verification: the enabled-demo e2e (four gated store pairs + the gated worker pair + baseline service accounts) runs at https://github.com/alienplatform/alien/actions/runs/30128396618 on these exact commits.

alongubkin:
@greptile-apps

ItamarZand88:
@greptile-apps review

PR #206, opened 2026-07-25
chore(sdk): add lease renewal endpoint

alongubkin:
## Summary

- regenerate the public API schema with the lease-only renewal route
- keep the Rust OpenAPI inputs synchronized

## Validation

- generated through the API schema copy command
- compiled the typed client through the manager integration

Tracking: ALIEN-322


alongubkin:
@greptileai review

PR #209, opened 2026-07-26
fix(ci): respect protected main in stable releases

alongubkin:
## Summary

- split stable releases into protected `prepare` and `publish` phases
- create an idempotent version-only release PR instead of pushing to `main`
- validate and tag only the merged release commit before publishing artifacts
- gate every downstream publication job on the publish phase

## Validation

- `actionlint -shellcheck= -ignore 'label .* is unknown' .github/workflows/release.yml`\n- `git diff --check`\n- locally exercised version detection against `v2.1.6`; correctly selects the `3.0.0` major bump\n\n[ALIEN-340](https://linear.app/alienplatform/issue/ALIEN-340/make-stable-releases-respect-protected-main)


alongubkin:
@greptile review

PR #214, opened 2026-07-26
fix(ci): qualify artifacts before stable release

alongubkin:
## Summary

- qualify stable release artifacts on the version-only pull request and record the exact source tree, artifact checksums, and image manifest digests
- publish only the qualified artifacts, with idempotent matching at crates.io, npm, GHCR, S3, GitHub Releases, and Homebrew
- delay the immutable tag and GitHub release until registries, images, and binary storage have accepted the release
- document prepare, publish, retry, and requalification operations

## Why

External registries cannot participate in a shared transaction. The release process therefore needs to be resumable: an absent artifact is promoted, an identical artifact is accepted, and a same-version conflict fails. Builds happen before merge, while publication promotes the exact qualified bytes after merge.

## Validation

- `actionlint -ignore 'label .* is unknown' .github/workflows/release.yml .github/workflows/release-qualification.yml`
- `bash -n scripts/build-npm-packages.sh`
- `git diff --check`
- qualified all five CLI npm tarballs from fixture binaries
- verified `publish-existing` accepts the five actual npm v3.0.0 tarballs
- verified `publish-existing` rejects a same-version tarball with different contents

Linear: ALIEN-343


alongubkin:
@greptileai review

alongubkin:
@greptileai review

alongubkin:
@greptileai review

PR #282, opened 2026-08-02
feat(cli): alien operations publish/list custom plugin bundles (ALIEN-428)

ab-alien-dev:
## What

Adds the customer-facing CLI to manage operations plugins. Operations plugins run inside a deployment **via the commands interface** (`plugin/operation`); this lets a customer publish a custom plugin bundle so its operations become available in their workspace.

## Commands

- `alien operations publish <bundle.zip>` — reads the bundle's `metadata.json`, validates it (name/version present, valid ZIP), and uploads the bundle to the platform (`POST /v1/operations/plugins`).
- `alien operations list` — shows the catalog (builtin + custom).

## Notes

- **Platform-gated** (`#[cfg(feature = "platform")]`): talks to the platform API via the authenticated http client, exactly like `alien managers`.
- Reads the ZIP with the `zip` crate and forwards the full `metadata.json` **verbatim** — the platform re-validates it authoritatively (the CLI doesn't depend on the private plugin crate, keeping the OSS boundary clean).
- base64-inlines the bundle so there's no multipart handling.

## Test plan

- [x] `cargo test -p alien-cli --features platform operations::` — **3 tests pass** (reads metadata, rejects missing `metadata.json`, rejects metadata missing required fields)
- [x] `cargo build -p alien-cli --features platform`
- [x] `alien operations --help` renders the command group

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -202,6 +202,7 @@ uuid = "1"
 hostname = "0.4"
 walkdir = "2"
 workspace_root = "0.1"
+zip = { version = "2", default-features = false, features = ["deflate"] }
 pem = "3.0"
 oci-tar-builder = "0.4"
 sha2 = "0.10"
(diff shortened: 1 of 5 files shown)


ab-alien-dev:
Reopening #281 on the same branch instead (per request).

PR #287, opened 2026-08-02
chore: release v3.3.2

alongubkin:
Version-only release PR generated by the stable release workflow. The **Release qualification** check builds and records the exact artifact set. After every required check passes and this merges, run the stable **publish** action from `main` to promote those qualified artifacts.


alongubkin:
Qualification found an unrelated release workflow failure after successfully building the SDK tarballs. Closing this generated version PR; it will be recreated from corrected main after #288 merges.

PR #293, opened 2026-08-02
fix: preserve setup authority in Platform client

alongubkin:
## Summary

- refresh the Rust Platform API schemas from the generated Platform OpenAPI source
- include `initialSetupAuthority` in the generated Rust client so durable deployment state round-trips losslessly

## Validation

- `cargo check -p alien-platform-api`
- inspected generated `codegen.rs` and confirmed `initial_setup_authority` is present

The public TypeScript SDK is intentionally not regenerated in this patch; Platform consumes the Rust build-time schemas for durable state conversion.


alongubkin:
@greptile-apps please review this schema-sync patch.

PR #296, opened 2026-08-02
fix: restore application access controllers

alongubkin:
## What changed

Register the standalone AWS, GCP, and Azure application-access controllers in the durable controller deserializer. The setup executor persists these controllers after each state-machine step, so every controller type created by direct setup must be restorable on the following tick.

This fixes the real AWS direct-setup failure reported as `RESOURCE_STATE_SERIALIZATION_FAILED: unknown controller type: AwsRemoteBindingsController` and prevents the equivalent GCP/Azure failures.

## Validation

- `cargo test -p alien-infra --all-features remote_bindings_controller_state_round_trips`
  - AWS round trip passed
  - GCP round trip passed
  - Azure round trip passed

The test serializes each controller through the production type tag and restores it through `deserialize_controller`, which is the failing persistence boundary.

### DIFF
--- a/crates/alien-infra/src/core/controller.rs
+++ b/crates/alien-infra/src/core/controller.rs
@@ -1117,6 +1117,20 @@ fn deserialize_controller_by_tag(
             deser!(crate::remote_stack_management::TestRemoteStackManagementController)
         }
 
+        // Application access controllers
+        #[cfg(feature = "aws")]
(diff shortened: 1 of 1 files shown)


alongubkin:
@greptile-apps please review this PR. The production failure is a durable controller-state deserialization gap; please verify all new in-tree provider controllers are registered and the tests exercise the real persistence boundary.

PR #300, opened 2026-08-03
fix(aws): make CloudFormation policy migration quota-safe

alongubkin:
## Summary

- emit compacted role grants as customer-managed policies instead of a replacement inline policy
- keep CloudFormation updates below the transient inline-policy quota while legacy policies are still attached
- preserve exact statement compaction and fan dependencies out to every replacement policy
- fail generation before exceeding AWS’s ten managed-policy attachments per role

## Root cause

The final consolidated policy was below IAM’s 10,240-character inline-policy quota. During an update from an older template, however, CloudFormation attached the new consolidated inline policy before deleting the legacy per-resource inline policies. IAM counts both sets during that transition, so the update failed even though the final state was valid.

Managed policies use a separate quota, allowing the new grants to coexist with legacy inline policies during migration. CloudFormation can then remove the old policies without a quota deadlock.

## Validation

- `cargo test -p alien-cloudformation` (92 passed)
- full rendered CloudFormation snapshot updated and reviewed
- 12-resource permission fixture produces no consolidated inline policies, stays within the ten managed-policy attachment limit, and preserves every resource grant

## Production reproduction

An update of an existing permission-heavy AWS stack failed while creating the replacement role policy with IAM `LimitExceeded` / maximum inline policy size. This change targets that update path; after release, the same stack will be retried end to end.

### DIFF
--- a/crates/alien-cloudformation/src/emitters/aws/helpers.rs
+++ b/crates/alien-cloudformation/src/emitters/aws/helpers.rs
@@ -50,7 +50,10 @@ pub(crate) fn chunk_managed_policy_statements(
             continue;
         }
 
-        if current.is_empty() {
+        if current.is_empty()
+            || managed_policy_document_size(std::slice::from_ref(&statement))?
+                > MAX_MANAGED_POLICY_BYTES
+        {
             return Err(AlienError::new(ErrorData::GenericError {
                 message: "AWS IAM statement is too large for a managed policy".to_string(),
             }));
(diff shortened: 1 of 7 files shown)


alongubkin:
@greptile-apps please review this quota-safe CloudFormation migration fix.

alongubkin:
@greptile-apps Both findings were valid and are fixed in 1fed5967f: the generator now counts existing AWS managed-policy attachments per role, reuses the established 5,500-byte managed-policy chunker, rejects oversized single statements, and has behavioral tests for both failure modes. Please re-review.

alongubkin:
@greptile-apps The later-oversized-statement bypass is fixed in the latest commit, with a direct small-then-oversized regression test. Please re-review.

alongubkin:
@greptile-apps review the latest commit 8997564cc. The previous finding is fixed and covered by a direct regression test.

PR #302, opened 2026-08-03
fix: quote CloudFormation mode values

alongubkin:
## Summary
- quote YAML 1.1 boolean-like `on`/`off` strings in all CloudFormation mapping values, not only parameter defaults and lists
- prevent custom resource properties such as `StackSettings.heartbeats` and `StackSettings.telemetry` from reaching registration as booleans

## Why
A real generated BYO-Bucket stack created its AWS resources but setup registration failed with HTTP 422 because CloudFormation interpreted unquoted `on`/`off` property values as booleans.

## Validation
- `cargo test -p alien-cloudformation quotes_yaml_1_1_boolean_like_strings_in_object_values`
- production-generated failure reproduced against AWS CloudFormation; after release, the same full stack will be recreated and exercised through an S3 write/read/delete round trip.

### DIFF
--- a/crates/alien-cloudformation/src/generator.rs
+++ b/crates/alien-cloudformation/src/generator.rs
@@ -657,17 +657,23 @@ fn quote_yaml_1_1_mode_scalars(yaml: &str) -> String {
     for line in yaml.lines() {
         let trimmed = line.trim_start();
         let indent = &line[..line.len() - trimmed.len()];
-        let replacement = match trimmed {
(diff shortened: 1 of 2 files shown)


alongubkin:
@greptile-apps please review this PR. This fixes a production-reproduced CloudFormation YAML 1.1 coercion in setup registration.

alongubkin:
@greptile-apps please review the final commit; it only accepts the expected quoted-value product snapshot.

PR #304, opened 2026-08-03
fix: skip empty vault sync without a vault

alongubkin:
## Summary
- skip the legacy command-token cleanup when a deployment has no secrets vault and owns no vault secrets
- apply the same rule during deployment deletion
- continue to fail closed if actual secrets require persistence but the vault is missing

## Why
A real resource-access-only AWS deployment imported its bucket and both scoped roles successfully, then failed provisioning because the empty secrets reconcile synthesized one legacy command-token deletion and tried to load a `secrets` vault that does not exist. Resource-access-only stacks intentionally do not need that vault.

## Validation
- `cargo test -p alien-deployment deployment_without_secrets_vault --quiet`
- production AWS handoff reproduced the failure with `0 desired and 1 removed` secret; after release the same deployment will be retried through CloudFormation and exercised with an object write/read/delete round trip.


alongubkin:
@greptile-apps please review this production-reproduced resource-access-only fix.

PR #306, opened 2026-08-03
chore: release v3.3.9

alongubkin:
Version-only release PR generated by the stable release workflow. The **Release qualification** check builds and records the exact artifact set. After every required check passes and this merges, run the stable **publish** action from `main` to promote those qualified artifacts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,7 +85,7 @@ dependencies = [
 
 [[package]]
 name = "alien-ai-gateway"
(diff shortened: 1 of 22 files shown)


alongubkin:
Closing this release preparation because the generated core schemas were stale. After #307 merges, a fresh v3.3.9 release PR should be prepared from the corrected main branch so qualification and publication use the same source tree.

PR #390, opened 2026-08-10
feat(sandbox): add the in-sandbox agent and the image bundle it ships in

ItamarZand88:
## Summary

Adds the agent that runs inside a sandbox and executes a caller's code, plus the image bundle it ships in. Layer 1 added the `Sandbox` resource type; this adds the process that actually runs something inside one.

When a caller asks a sandbox to run a command:

1. The agent authorises the request — a signed capability, or the transport itself where the cloud already scopes the caller to one sandbox.
2. It resolves any path the request names **through the kernel**, so a path that would leave the session root is refused rather than checked and then opened. **This is the heart of the change.**
3. It spawns the command with a fresh environment, as an unprivileged user, in its own process group, and streams stdout and stderr back as they are produced.
4. The deadline kills that whole group, so what the command forked goes with it.

This turns the sandbox from a resource you can declare into one you can run code in.

## What I did

The agent is a small HTTP server that ships inside the image. It runs as root so it can drop to an unprivileged user before every spawn; inside a MicroVM that drop sits behind hardware virtualisation, which is the tenant boundary.

Two decisions are worth pointing at:

- **Paths go through `openat2`, not a resolver.** Resolving a path and then opening it by name leaves every check in a race window, and the code being confined runs in the same guest and can drive both sides of it. `RESOLVE_BENEATH | RESOLVE_NO_SYMLINKS` makes resolution and open one operation. This removed the hand-written component walk, the canonicalisation, and the containment check rather than adding a fourth guard next to them.
- **`spawn` and `spawn_sandboxed` are two functions, not a flag.** One inherits the environment for our own helper processes; the other clears it. The agent's environment names its own port and session, so handing it to a caller's code hands over a map to the API running it.

## Files touched

- `crates/alien-sandbox-agent/` — the agent: HTTP surface, exec, file transfer, path confinement, privilege drop, PID-namespace support for runtimes that grant it.
- `crates/alien-core/src/sandbox_process.rs` — process framing, shared with the GCP launcher path.
- `crates/alien-build/src/sandbox_bundle.rs` — the image bundle and its Dockerfile.
- `crates/alien-bindings/src/error.rs` — the error variants the agent's callers see.

## How I tested

- 27 unit + 14 protocol tests, run **on Linux in a container** as well as on macOS. The Linux run is the one that counts: it exercises the real `openat2` path, and it caught a traversal on write returning 500 instead of 400 that macOS could not have surfaced.
- A standalone probe against a real kernel: a legitimate file opens; a symlink to an outside file, a symlinked parent directory, a dangling symlink on write, `..` traversal and an absolute path are all refused. The control arm reads the same symlink successfully without confinement, so the refusals are the code working rather than a broken fixture.
- Each security-relevant test was mutation-tested — the fix reverted, the test observed to fail, the fix restored. That covers the dangling-symlink escape, environment inheritance, the process-group kill, the refusal of the agent's own supervised code, and Dockerfile directive injection.
- A standalone probe confirming a connecting socket can be attributed to its owning user before that was relied on: a connection from the sandbox user is identified as such, and one from another user is not mistaken for it.
- `protected_hardlinks` checked on the actual MicroVM base image, as the sandbox uid, with a control arm.
- `cargo check` for `alien-build` and `alien-bindings` under their individual feature combinations, not just the workspace build.

Security review of this diff, since it runs untrusted code:

- A caller planting a symlink — refused by the kernel at open, verified against a real kernel.
- A caller racing a rename between check and open — no window; resolution and open are one call.
- A caller reaching the agent's own API from inside the guest — refused. The agent does not serve a request whose connecting socket belongs to the identity it runs commands as, so the command it started cannot ask it for more work. Verified against a real kernel, with a control arm showing a caller on loopback under a different user is unaffected — a proxy terminating inside the guest still works.
- An image reference carrying a newline — refused before it renders, since a reference reaching a generated Dockerfile would otherwise write its own directives. The check sits inside the renderer, so neither entry point can skip it.
- A command outliving its deadline — the process group is killed; a `setsid` child still escapes, which needs a cgroup and is called out in the module docs.
- A setuid binary in a caller-supplied base image — `no_new_privs` is set after the drop.
- Supplementary groups surviving the drop — shed before the uid changes.
- Hard links — deliberately not addressed here and documented; a link is the inode, so no resolver can tell it apart, and `protected_hardlinks` bounds it to files the caller could already write.

Nothing turned up.

### DIFF
--- a/.github/workflows/ci-fast.yml
+++ b/.github/workflows/ci-fast.yml
@@ -246,6 +246,19 @@ jobs:
         run: |
           cargo test -p alien-manager --features openapi --lib api::tests::openapi_declares_http_bearer_security_without_dropping_schemas
 
+      - name: Sandbox agent privileged tests
+        # The agent runs as root and drops to a different uid before every spawn.
+        # The sandbox's file and working-directory handling turns on those two
+        # identities differing, which nothing running under a single uid can
+        # observe, so these are `#[ignore]`d and need root. Built as the runner
+        # and only the test binary re-run under sudo: a root `cargo` would leave
+        # root-owned artifacts behind for the cache upload.
+        run: |
+          binary=$(cargo test -p alien-sandbox-agent --test privileged --no-run --message-format=json \
+            | sed -n 's/.*"executable":"\([^"]*privileged-[^"]*\)".*/\1/p' | head -1)
+          test -x "$binary"
+          sudo "$binary" --ignored
+
       - name: Remote bindings tests (non-cloud)
         # `platform-sdk` is not a default feature, so `remote::` is cfg'd out of
         # the workspace run. The cloud client crates' unit tests already run
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1038,7 +1038,7 @@ dependencies = [
 
 [[package]]
 name = "alien-sandbox-agent"
-version = "3.3.11"
+version = "3.3.13"
 dependencies = [
  "alien-core",
  "alien-error",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1038,7 +1038,7 @@ dependencies = [
 
 [[package]]
 name = "alien-sandbox-agent"
-version = "3.3.13"
+version = "3.3.14"
 dependencies = [
  "alien-core",
  "alien-error",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -319,6 +319,7 @@ dependencies = [
  "url",
  "which 6.0.3",
  "workspace_root",
+ "zip 2.4.2",
 ]
 
 [[package]]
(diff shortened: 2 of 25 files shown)


ItamarZand88:
@greptile-apps one commit added, `b56142ba9`. It came from an intermittent CI failure on the top of the stack, and it is a real hole in a security control rather than a flaky test.

`transport_authorization_refuses_the_code_the_agent_runs` occasionally returned **204** where **403** is required — the agent served the very code it is meant to refuse.

**Mechanism, reproduced on Linux as a non-root user (not inferred):**

`find_owner` reads `/proc/net/tcp` to attribute a connecting socket to a uid, matching a row on address + port and taking the first hit. But a closed socket stays listed in `TIME_WAIT` (`st=06`) **under uid 0**, on the same address:port, and the kernel hands that port to a new connection while the stale row is still in the table:

```
stale TIME_WAIT on client port 49620 -> [(st 06, uid 0)]
kernel REUSED port 49620 on connect #7099 -> rows: [(06, 0), (01, 1234)]
```

The stale uid-0 row sorts first, so the live uid-1234 caller was attributed to root; `0 != exec_uid` → `decide()` served it. On a CI runner with eleven parallel tests opening loopback connections, that collision is exactly what fired.

**Fix:** only an established row (`st == 01`) is read — only a live socket has an owner. The new unit test lays a `TIME_WAIT` row ahead of the live one on the same port and asserts the live owner; without the check it returns `Some(0)`, i.e. exactly what CI saw. Verified by removing the check and watching that test go red.

Worth noting the existing doc-string on `decide()` already named the cousin of this hazard — *"a command can write its request and close before the socket table is read"* — and fails closed for it. This is the other half: a *previous* socket's close being read as the current caller.

PR #401, opened 2026-08-11
feat(sandbox): add the Local and Kubernetes backends and the cloud bindings

ItamarZand88:
## Summary

Adds the code that actually runs a sandbox session: the Local and Kubernetes backends, and the bindings an application calls on AWS, Azure and GCP. Layer 4 rendered the infrastructure; this makes a session start, run a command, and stop.

When an application asks a Kubernetes sandbox to run something:

1. It presents the ServiceAccount token Kubernetes already mounted in its pod to the session broker.
2. The broker checks that token with a `TokenReview`, takes a warm pod from the pool, and labels it with the session.
3. **It mints a capability signed with a key the application never sees, derived from the sandbox's own id rather than named by the caller.**
4. The application drives the session over the agent port, presenting that capability on every call.

## What I did

Three decisions are worth pointing at, each of which came out of a review finding rather than a plan:

- **The broker derives the signing key; it does not accept one.** The claim request used to carry the secret name. A workload holding one sandbox's handle could name a sibling's secret and get a capability minted under that sibling's key. The name is deterministic, so the field is gone from the wire format.
- **Azure refuses what it cannot enforce.** The Azure data plane takes no egress policy and no per-session ceiling, and the binding carries neither — so a declared `deny` was accepted and dropped, leaving a stack reading as restricted while the code ran with open egress. Azure now publishes `false` for both capabilities, which turns the declaration into a plan-time refusal. `allow` is still accepted: it asks for no restriction, so a backend that ignores it fails loudly on the first blocked connection rather than quietly under-protecting.
- **Terminate reports failure.** Both the Local and Azure paths discarded the result of their delete and returned success. Terminate is the containment kill switch — reporting success for a daemon error or a throttle tells the caller untrusted code has stopped when it has not. Only an absent session is success now.

A fourth is smaller but had the same cause: the Azure `get` treated any error whose rendered text contained `404` as "session gone", and the data plane formats the whole response body into that text. A throttle, or a path containing `404`, read as gone — which starts a second sandbox while the first keeps running. The status is carried structurally now.

## Files touched

- `crates/alien-infra/src/sandbox/` — the Kubernetes controller, warm pool, session broker and its route.
- `crates/alien-local/` — the Docker-backed local manager and its loopback route.
- `crates/alien-bindings/src/providers/sandbox/` — the five backends behind one trait.
- `crates/alien-k8s-clients/`, `crates/alien-aws-clients/`, `crates/alien-azure-clients/` — the API surfaces those call.

## How I tested

- `cargo test` across the touched crates — the Kubernetes broker, warm pool, path confinement, capability minting and the five providers.
- Mutation-tested every fix above by breaking it and watching a test fail: flipping Azure's egress capability back to `true`, removing the chart's ingress rule, and reverting the `404` classifier to substring matching. Each fix has a test that fails without it.
- Live suites (`#[ignore]`d, run by hand): a real GKE Autopilot cluster for the Kubernetes backend, and Docker for the Local one.

Security review of this diff, since it mints the credential that reaches inside a session:

- A workload minting a capability under another sandbox's key — the broker derives the key name from the sandbox id, and a request naming one no longer deserializes (`kubernetes_route.rs`).
- A capability replayed against another session — the AWS token is scoped to its MicroVM by the platform, and the Kubernetes capability is bound to the claimed session and checked at the agent.
- A declared egress restriction silently dropped — refused at plan time on any backend that cannot enforce it, with the capability table as the single source of that answer.
- Untrusted code left running after terminate — every delete failure now propagates except an absent session.
- A rejected ServiceAccount token accepted — the apiserver answers 200 with `authenticated: false`, so the verdict is read rather than the status code, and namespace matching is a full-prefix check.

One thing this diff does not close, stated plainly rather than left to a reader: the broker authenticates the caller but does not check *which* sandbox it may claim from. `claim` verifies a ServiceAccount in the deployment's namespace and then takes the `sandboxId` from the request. Binding an authenticated identity to an entitled set of sandboxes needs state the broker does not have yet.


ItamarZand88:
Both remaining findings are correct-as-designed rather than defects — here's the evidence for each.

**AWS session ownership at image scope.** The image is one per *declared sandbox*, not a shared namespace: the emitter binds `imageArn` to the ARN AWS assigns to the single image resource it emits for that one sandbox (`crates/alien-terraform/src/emitters/aws/sandbox.rs`), and the binding reads that ARN back off the resource. Two bindings that resolve to one image are two bindings on one declared sandbox — the sharing the declaration asked for. Ownership is now proven per-session by an ARN-scoped `GetMicrovm` comparing the session's own `imageArn` against the bound image (`owned_microvm`, `crates/alien-bindings/src/providers/sandbox/aws.rs`), so a sibling *sandbox's* session id is refused before any token is minted.

**Azure command returns after its deadline.** This is forced by the API, not a choice:
- The ADC data-plane `executeShellCommand` has no server-side timeout — the request body is `command` plus an optional `workingDirectory` and nothing else (read out of the preview SDK). The `timeoutInSeconds` that looks relevant belongs to `sessionPools` (`dynamicsessions.io`), a different product; Sandboxes are `azuredevcompute.io`/`SandboxGroups`.
- `delete` is accept-then-poll async — "returns before it is gone; confirm by polling `get_sandbox` to 404" (`crates/alien-azure-clients/src/azure/sandbox_data_plane.rs:66`).

So at the instant the deadline fires, "the command is contained" is not yet a knowable fact. Returning `deadlineExceeded` then would assert containment the code never checked. The deadline bounds the *command*; the call returns once containment is confirmed (bounded at deadline + terminate-confirm window), and it reports `SandboxUnreachable` / "may still be running" rather than a false `deadlineExceeded` when it cannot confirm. A punctual return and a guaranteed-containment claim can't both hold on an accept-then-poll delete API — the design picks the honest guarantee.

@greptile-apps please re-review.

ItamarZand88:
@greptile-apps review

Both points in the last review look like misreads of deliberate choices, so I have not changed the code. Reasoning, in case it is the framing that is off rather than the judgement:

**Image-ARN equality as the ownership check** (`sandbox/aws.rs`) — the sandbox, not the individual session, is the trust boundary. Each Sandbox resource builds its own image, so the ARN identifies the resource; sessions sharing one image are sessions of one workload holding one binding, and there is no second tenant on the other side of that check to cross over to. A session id from a different sandbox carries a different image ARN and is refused. `owned_microvm` also refuses an absent `imageArn` rather than assuming a match, so a response the client failed to parse cannot pass.

**Awaiting cleanup before returning the deadline error** (`sandbox/azure.rs`, `sandbox/local.rs`) — this is the point of the branch, not an oversight. Azure`s data plane takes `command` and an optional `workingDirectory` and nothing else, so there is no server-side timeout to ask for and ending the session is the only lever that stops an overrun. Returning `deadlineExceeded` before that containment is confirmed would assert something the code had not yet made true. The deadline bounds the untrusted code, not the caller`s wall clock, and the rationale is recorded at the call site.

If either is still judged a real defect, the specific crossing or the contract the delay violates would help, since the current summary does not name one.

ItamarZand88:
@greptile-apps please re-review — the two open points are settled below, and the reasoning is now written into the code rather than left implicit.

## `deadlineExceeded` means the command has stopped

Both readings have been suggested across reviews — return at the deadline without confirming containment, or confirm containment and land late. They cannot both hold, so this follows the resource's own rule: a deadline always terminates the command, and there is no unbounded exec. Reporting a stop that was only *requested* would leave untrusted code running while the caller is told it was contained — the one claim this resource exists to make.

Every backend now reports it the same way:

1. AWS and Kubernetes: the agent kills the process group and returns after the kill.
2. Azure and Local: the command runs under `timeout -s KILL <deadline>` **inside the session**, so it is killed at the deadline, its output is kept, the stream ends in `deadlineExceeded`, and the session stays usable. The call lands at the deadline.
3. Only a session image that cannot run `timeout` at all reaches the client-side guard, and there the session itself is ended and the call returns once it is confirmed gone.

Step 2 is the one that changed since the earlier reviews: the overrun is contained *in place* rather than by ending the session, so the "late" path is now reachable only by an image without `timeout` — bounded by the deadline, the grace, and the delete's confirmation window. `RunCommandRequest::deadline`, `AzureSandbox::execute_within` and `LocalSandbox::exec_within` all state this now.

Related, from earlier rounds and already in this PR: the `timeout` wrapper is chosen by a one-time side-effect-free probe (`command -v timeout`) cached per provider, so a caller's command is never executed twice; the bound carries milliseconds; a sub-millisecond deadline is refused rather than rounded up.

## AWS session ownership is the image ARN, and there is one image per sandbox

`ensure_owned` / `owned_microvm` compare the session's `imageArn` against this binding's own, which is not a naming convention: the emitter emits exactly one `AWS::Lambda::MicrovmImage` per declared `alien.Sandbox` and binds `imageArn` to the ARN AWS assigned to *that* resource, read back off it (`crates/alien-terraform/src/emitters/aws/sandbox.rs`). Two declared sandboxes therefore cannot share an ARN however their names collide. Two bindings that do resolve to one image are two bindings of the *same* declared sandbox — that is the sharing the declaration asked for, and IAM cannot answer it because the stack binding scopes the mint to `microvm-image:<stack prefix>-*`, which matches every sibling. That is exactly why the check reads the session rather than trusting IAM.

ItamarZand88:
@greptile-apps two more fixes just landed on this branch, both from your last pass:

- **`124` alone no longer means the deadline.** It is an ordinary exit status, so the wrapper's kill is now told apart by the clock — the run must have reached its deadline (`wrapper_killed` in `providers/sandbox/mod.rs`). A command that returns 124 well inside its deadline is reported as a plain `Exit { code: 124 }` on both Azure and Local, each with a test.
- **The guard's arithmetic saturates.** `deadline + DEADLINE_GRACE` panicked for a deadline near `Duration::MAX`; both providers now use `saturating_add`.

ItamarZand88:
@greptile-apps re-review please — the deadline classification is no longer inferred at all, and the last three findings are fixed at the root rather than guarded.

**The wrapper does the killing and reports it.** Neither the exit status nor the clock can tell a deadline kill from a command's own ending, and both were measured rather than assumed: `timeout -s KILL` exits **137** on GNU coreutils 9.1 and BusyBox alike; `timeout -k` exits **124** on GNU but **143** on BusyBox, and BusyBox reports **0** for a command that traps `TERM`; and what a provider measures is its own round trip, transport included. So the command now runs under a small `sh` program that starts a killer, and the killer prints its report only when its signal was delivered.

**The report cannot be forged.** The nonce is drawn *inside the session* from `/dev/urandom`, announced on the first line of stderr and repeated only on a kill. It is in neither the parent's command line nor its environment — verified by running a command that reads `/proc/$PPID/cmdline` and `/proc/$PPID/environ` inside both an Alpine and a Debian container: the program text is visible, the nonce is not. A command cannot read what has already been written to the stream it inherits either.

**Local stderr is assembled before it is read**, so an announcement and a repeat that arrive in different Docker frames are still one report; a test pins exactly that case.

Two things fall out of this rather than being added: `timeout` is no longer required at all — only `sh` and `/dev/urandom` — so the probe and the image-without-`timeout` fallback are gone, and with them the last path that could report a deadline late.

Also in this branch since the last pass: a deadline is refused when it is below a millisecond or beyond a documented 24-hour ceiling, so no unrepresentable timer deadline is reachable (this replaces the earlier saturating add, which avoided the arithmetic overflow but still handed Tokio a duration with no representable instant).

The local `greptile review` on this layer now reports 5/5 with no comments.

ItamarZand88:
@greptile-apps two changes since the last pass, and one boundary I want to state plainly rather than keep re-cutting.

**Descendants are now killed with the command.** The wrapper starts it with `setsid`, so the deadline signal goes to the whole process group, not one pid. Measured in Alpine (BusyBox) and Debian (GNU coreutils 9.1): a command that spawns `sleep 30 &` and then sleeps leaves **zero** survivors after the deadline, while a fast command is untouched and gets no report.

**An image that cannot do that runs nothing.** There is no weaker fallback: without `setsid` (or `/dev/urandom`) the wrapper exits 126 before starting the caller's command, and the provider returns a typed failure saying so. Verified with a PATH holding `sh`, `od`, `tr`, `sleep` and no `setsid`: exit 126, no announcement, nothing executed. A deadline that cannot be enforced is refused, not approximated.

**On the remaining case — a descendant that calls `setsid` itself:** no signal sent from inside the session reaches it, and that is not specific to this wrapper. The agent-backed backends bound the same way — `libc::kill(-(pid), SIGKILL)` in `alien_core::sandbox_process::kill_all` — so the process group is where this codebase draws the line everywhere. What bounds a deliberately detached process is the sandbox session itself: `terminate`, or the session's own lifetime ceiling, which is part of the resource's contract. In-session PID-namespace isolation is not available on these two backends (the agent's own comment records that no runtime grants the capability an unprivileged drop would need).

So rather than claim containment the design does not have, the guarantee is now written where callers read it — on `RunCommandRequest::deadline` and on the wrapper: *the command and its process group stop at the deadline; a process that detaches itself is bounded by the session*. If you think that line belongs somewhere else, that is a design call worth making explicitly rather than by wording — but it is the same line every other sandbox backend in this repository draws.

ItamarZand88:
@greptile-apps one commit added here (`feba9cf`), found while chasing a session-leak path on the e2e layer above.

The Local sandbox delete flow reaped every session and *then* removed the route. The route is what accepts session creates, so the window between the sweep and its removal is one where a create can still land — leaving a container behind for a sandbox that no longer exists, which is exactly what the route removal was there to prevent. The existing comment named that hazard and the ordering left it open.

The route now closes first, which is what makes the sweep final: nothing can be created once it is gone.

Safe to swap: `reap` finds containers by Docker label (`sandbox_manager.rs` `list_sessions`) and `SandboxRoute::remove` only touches the in-process route map, so neither depends on the other.

No test: asserting call order through a mock would pin the implementation rather than the behaviour, and there is no seam here that observes the window without inventing one. The reason is stated at the call site instead.

ItamarZand88:
@greptile-apps review please — this PR has had three pushes since your last review (`feba9cf` was the last commit reviewed; head is now `5b61ecde5`) and the review check never re-ran. Since then it gained: the route-before-sweep ordering fix on Local delete, and `5b61ecde5` bounding how long an in-sandbox agent may take to answer (headers only, via `tokio::time::timeout` around `send()`, deliberately not reqwest's body-inclusive `.timeout()`).

ItamarZand88:
@greptile-apps to your open question — confirmed from `crates/alien-local/src/sandbox_route.rs`: `SandboxRoute::remove` is infallible, and by signature (`pub async fn remove(sandbox: &str)` returns `()`), not by convention.

Every early return is the desired end state rather than a failure:
- `ROUTES` never initialised → no route ever served, nothing to close.
- no map entry for the sandbox → the route is already gone.
- stopping the listener is `drop(route.shutdown.take())` — dropping a oneshot sender is what the listener's graceful shutdown waits on, and a drop cannot fail.

The one fallible call is deleting the token file, and it is deliberately best-effort (`let _ = remove_file`), because an already-removed file is exactly the goal. There is no path on which the route is still serving and the caller is not told — the "route still serving" outcome does not exist for this function to swallow.

So the asymmetry you noticed (`remove` discarded, `reap` propagated) is the right one: `reap` talks to Docker and can genuinely fail with containers left running, which the caller must hear; `remove` cannot.

ItamarZand88:
@greptile-apps re-review please. Both P1s were correct, and both are fixed — with two more things found in the course of verifying them.

## 1. Stalled `run_command` no longer invites a retry (`c1e2d94`)

You were right: the timeout branch returned retryable `SandboxUnreachable` for every operation, and for `/v1/exec` the command may already be running. Now a stalled `run_command` is refused as `SandboxCommandFailed { failure: "outcomeUnknown" }` — non-retryable, outcome stated as unknown — while file operations stay retryable. `outcomeUnknown` is the code this file already uses for the same meaning on the body-read path, so a caller branching on `failure` sees one code for one condition. Two tests pin the two branches by the error's own `retryable` flag; the run_command one fails when the branch is removed.

Also fixed the timing assertion in `an_agent_that_never_answers…`, which was flaking 2 in 6 under the full suite: it measured from before the listener was even bound, against a 1s multiple. It now bounds against the agent's own answer time (the property), and still fails at 4.76s when the timeout is removed.

## 2. The deadline killer finishes its report before it is reaped (`c5c0185`)

Confirmed the race and reproduced it. The natural window is too narrow to observe on an idle host — **0 misses in 600 runs** — so I did not claim reproduction from that. Widening the killer's kill→printf gap by 20ms turned **100 of 100** deadline kills into misreports on both BusyBox and dash, which is what a loaded scheduler does for free. The parent now branches on the exit: 137 → `wait` the killer so its report lands; otherwise stop it. Same harness on the fixed program: **0 of 100** missed.

Two further findings from testing the fix, both on the **original** program too:

- **The stopped killer left its sleeper running** for the rest of the deadline — killing the subshell does not reach its `sleep` child. Verified 1 orphan on both shells for the original. The killer now reaps its own sleeper on the way out; verified 0.
- **A tempting fix leaks the nonce.** Making the killer a session leader by passing the nonce as an argument puts it in `/proc/<pid>/cmdline`, and a hostile command reads it — reproduced: the command extracted the nonce and could forge its own deadline. So the killer stays a subshell and the nonce reaches it only as a variable. Verified: a command scanning `/proc` finds nothing.

Full matrix on the exact emitted program, `alpine:3.20` and `debian:bookworm-slim`: 0/100 misreports under the widened window (original: 100/100); 200/200 deadline kills reported at natural timing; early exit keeps its real code, `killed:false`, 0 sleepers; an external SIGKILL before the deadline exits 137 and is correctly **not** called a deadline; hostile `/proc` scan finds no nonce.

ItamarZand88:
@greptile-apps re-review please. Of the three findings on `c5c0185`: one was a misread, two were correct, and both correct ones are fixed — one of them by making the wrapper simpler rather than more branched.

**Line 240 — not a defect.** That line is the non-`run_command` fallthrough, which returns retryable `SandboxUnreachable` for file operations on purpose (they are idempotent). The `run_command` branch immediately above it returns the non-retryable `outcomeUnknown`. Nothing to change there.

**Line 221 — correct, and it was my miss.** I fixed the *timeout* arm and left the *transport-error* arm: a connection that drops before headers hit the same indeterminacy for `run_command` — the agent may have taken the command before the socket went — and still mapped to retryable. Both no-answer paths now share one rule (`unanswered()`): `run_command` → `outcomeUnknown`, non-retryable; everything else retryable. New test drops the socket on accept and asserts the refusal is not retryable; it fails against the version that fixed only the timeout.

**mod.rs:85 — correct, and the fix removed the branch entirely.** You were right that an external SIGKILL before the deadline made the parent wait out the killer's full sleep. Rather than distinguish the two 137s with a pid check or a temp file (a hostile command reads `/proc`, and the session root is read-only — both were tried and rejected), the parent now always does the same thing: stop the killer, then await it. The killer's own state decides what that means: asleep, it reaps its sleeper and exits at once — an external kill returns immediately; past its sleep it ignores the stop (`trap '' TERM`) so its report lands before it exits. No branch on the exit code, no constant, and no way to mistake an external kill for the deadline, because the killer only ever reports a kill it made.

Verified on the exact emitted program in `alpine:3.20` and `debian:bookworm-slim`: 0/100 misreports under the widened window (original: 100/100); 200/200 deadline kills at natural timing; early exit keeps its code, 0 sleepers; **external SIGKILL now returns in 0s** (was: full deadline), `killed:false`, 0 sleepers; a stderr reader that has gone away does not hang the wrapper; hostile `/proc` scan finds no nonce.

ItamarZand88:
@greptile-apps one commit added, `09a7ffc91`, found by following the leak question on #403 all the way down to its root.

`SandboxRoute::remove` dropped the listener's shutdown sender and **returned** — a signal, not a fact. Graceful shutdown lets a request already inside the route finish, and nothing awaited the task that does so. So a create mid-flight at removal could commit *after* removal returned and after the reap that follows had already swept and found nothing: a session with no route left to reach it by. Removal now awaits the listener task; returning from it means no create is in flight, which is what makes the sweep after it trustworthy.

Two live-Docker tests pin what the e2e check now leans on. **A session nobody terminated is reaped with its sandbox** (1 → 0); it fails against a reap that counts but does not terminate. **A create parked inside the route when removal fires is visible to the reap that follows** — made genuinely slow by an image the daemon has to pull, since a warm one commits in milliseconds and would let removal win every race without proving anything. Against a removal that only signals, it fails 3 of 3 with `committed=true, reaped=0` — the exact interleaving.

The first draft of that second test passed both ways (0/5 detected) because it raced two futures on one runtime and the create won every time. I replaced it rather than ship a test that cannot fail.

ItamarZand88:
@greptile-apps re-review please. Correct again, and this one closes the rule out — `9f3472ca7`.

You found the third face of the same defect. I had made "no answer before headers" non-retryable for `run_command` (stall, and dropped connection), but `frame_stream` still refused a body that **failed or ended short of its terminal frame** as retryable `SandboxUnreachable`. That is the strongest outcome-unknown case of the three: the command had demonstrably started — frames were arriving — and only its end is lost. A retry re-executes it.

Both arms now refuse through the same `unanswered()` rule: `run_command` → `outcomeUnknown`, non-retryable. One helper, three faces (stall / dropped-before-headers / lost-after-headers), one vocabulary — and its doc now says what it covers.

The reason this survived every earlier pass is worth stating: the existing test asserted the bug. `a_stream_without_a_terminal_frame_is_a_transport_failure` literally said *"a dropped stream must stay retryable"*. It is now `…_is_an_unknown_outcome`, asserts `SANDBOX_COMMAND_FAILED` and `!retryable`, and fails against the retryable form — verified by swapping it back in.

ItamarZand88:
@greptile-apps re-review please. The finding at `mod.rs:84` — a command that completes normally at the deadline being misreported as killed because `kill` succeeds on the not-yet-reaped group — rests on the group still existing after the command exits. I tested that interleaving directly, in the wrapper's exact shape (`setsid "$@" & cpid=$!`), on `alpine:3.20` (BusyBox) and `debian:bookworm-slim` (dash):

**Command exits on its own; the parent has not yet `wait`ed:**
```
cpid=7  pgid-of-cpid=<gone>
kill -KILL -"$cpid" on the exited command: FAILED -> no nonce
```
Both shells. The command is the session leader and the only member of its group, so the group ceases to exist the moment it exits — there is no zombie group for `kill` to succeed against, and the nonce is not printed. The wrapper reports the real exit.

**For contrast, while the command is running** (the case the kill exists for):
```
kill -KILL -"$cpid" on the RUNNING group: SUCCEEDED; sleep is gone
```
So the kill reaches a live command and misses an exited one — exactly the discrimination the design relies on: the killer only reports a kill it made. This is the same property the widened-window proof on this PR established from the other side: 200/200 genuine kills reported, and an early exit reported as its real code.

If there is a concrete interleaving on Linux where a group with an exited leader and no other members still accepts a signal, I will fix it — but the probes above say the premise does not hold.

ItamarZand88:
@greptile-apps

The deadline bounds the untrusted code, not the caller's patience — and `deadlineExceeded` is a statement that the command **has stopped**, never that a stop was requested.

Returning at exactly the ceiling without confirming the kill would make that error a lie: the caller is told the command stopped while it is still running. That is the same defect accepted as real on #402 — a cancel that reports success while untrusted code keeps executing. Returning early here would reintroduce it at a different layer.

The path being described is a fallback, not the normal one. Every overrun is ended **in place, at the deadline**, by the in-session `timeout` wrapper (`local.rs`: *"the deadline is enforced inside the session: the wrapper kills the command"*). The client-side guard is reached only by a session that could not run `timeout` at all. On that path the return is bounded rather than early, and the bound is stated: the deadline, the grace, and the delete's confirmation window.

Both backends document exactly this, and deliberately agree with the agent-supervised ones, where the agent waits for its kill before reporting:

> This is the one path where untrusted code is known to be running past its deadline, so it is bounded rather than early: the deadline, the grace, and the delete's confirmation window, and it is reached only by a session that could not run `timeout` — every other overrun is ended in place, at the deadline.

Making the call return sooner would not shorten how long the untrusted command runs. It would only stop the caller from learning that it did.

ItamarZand88:
@greptile-apps please re-run the review — the deadline/termination concern is resolved in the thread above, and the last review run was cancelled rather than completed.

ItamarZand88:
@greptile-apps this finding was reviewed and withdrawn earlier in this thread — your own reply: *"The argument holds. Looking at `exec_within`, `self.terminate(session_id).await?` blocks before the ..."*

The reasoning stands unchanged: `deadlineExceeded` asserts the command **has stopped**. Returning at the ceiling without confirming the kill makes that a lie, which is the same defect accepted as real on #402. The path is a fallback reached only when the session could not run `timeout`; every other overrun is ended in place, at the deadline, and the return is bounded by deadline + grace + confirmation window.

The only change on this push is a fix to `mod.rs`, where the kill is now credited only when the exit status confirms it — a command that finished as the killer fired keeps its own result. Please re-run the review.

ItamarZand88:
@greptile-apps this finding was reviewed and withdrawn earlier today — your own reply on this thread: *"The argument holds. Looking at `exec_within`, `self.terminate(session_id).await?` blocks before the ..."*

`crates/alien-bindings/src/providers/sandbox/azure.rs` is byte-identical to the version you withdrew it against; this push did not touch it. The only change here is in `alien-core`: `files` becomes a published capability, because Azure's data plane has no file transfer and the old contract called it part of the guaranteed floor.

The reasoning stands unchanged: `deadlineExceeded` asserts the command **has stopped**, so returning before the kill is confirmed would make that a lie — the same defect accepted as real on #402. The path is a fallback reached only when the session could not run `timeout`; every other overrun is ended in place, at the deadline. Please re-run the review.

ItamarZand88:
@greptile-apps this finding was reviewed and withdrawn earlier today — your own reply on this thread: *"The argument holds. Looking at `exec_within`, `self.terminate(session_id).await?` blocks before the ..."*

Both flagged files are byte-identical to the version you withdrew it against. This push changed `providers/sandbox/gcp.rs` (an environment the launcher cannot carry is now refused instead of silently dropped, and a zero deadline is refused) and one error message in `alien-infra/src/sandbox/local.rs` — a different file from the `providers/sandbox/local.rs` named above.

The reasoning is unchanged: `deadlineExceeded` asserts the command **has stopped**. Returning before the kill is confirmed makes that a lie. The path is a fallback reached only when the session could not run `timeout`; every other overrun is ended in place, at the deadline. Please re-run the review.

PR #402, opened 2026-08-11
feat(sdk): expose the sandbox binding to TypeScript

ItamarZand88:
## Summary

Exposes the sandbox binding to TypeScript, so an application written in TypeScript can create a session, run a command, move files and stop it through the same contract the Rust bindings use.

When a TypeScript application asks for a sandbox binding:

1. The generated N-API layer hands back the same binding the Rust provider resolved.
2. The SDK wraps it in a typed surface — `create`, `runCommand`, `readFile`, `writeFiles`, `terminate`.
3. **A capability the platform does not support is a typed error naming both the platform and the capability, not a silent no-op.**

## What I did

The published capability set is the interesting part. A portable application cannot assume every backend does everything — preview exists only on AWS, hostname allowlists exist nowhere, and ceilings are enforced on some platforms and refused on others. Rather than let a call fail somewhere deep in a cloud SDK, the capability set is exposed to TypeScript as data, so an application can branch before it calls.

The types are generated from the Rust definitions rather than hand-written, so the two languages cannot drift: the binding JSON is the cross-language contract.

## Files touched

- `crates/alien-bindings-node/` — the N-API surface.
- `packages/bindings/src/types.ts` — the generated binding types.
- `packages/sdk/src/index.ts` — the public export.

## How I tested

- `packages/bindings/tests/sandbox.test.ts` — the binding shape and the capability set as TypeScript sees them.
- `pnpm generate` — confirms the committed types match the Rust definitions; CI fails the build if they drift.

## What changed since the last review

The Node binding now advertises `files`, so a TypeScript caller can branch on it before calling
`readFile`, `writeFiles` or `mkdir`.

1. `capabilities()` destructures the capability set by field name, with no `..`.
2. **The new field had to be named there or the crate would not compile** — which is why the
   destructure is written that way, and is what caught it.
3. `"files"` joins the returned list on every backend that supports it.

The three file operations now document the capability they need, matching how `suspend` and
`resume` already document `suspendResume`.

### Files touched

- `crates/alien-bindings-node/src/sandbox.rs` — `files` in the destructure and the advertised list
- `packages/bindings/src/types.ts` — the capability requirement on the three operations

### How I tested

- `cargo test -p alien-bindings-node` — 28 pass
- `npx vitest run src/__tests__/factories.test.ts` — 40 pass
- Checked out an intermediate commit of this branch and ran `cargo check -p alien-bindings-node`
  to confirm every commit compiles, not just the tip

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -273,6 +273,7 @@ name = "alien-bindings-node"
 version = "3.3.14"
 dependencies = [
  "alien-bindings",
+ "alien-core",
  "alien-error",
  "chrono",
  "futures",
--- a/crates/alien-bindings-node/Cargo.toml
+++ b/crates/alien-bindings-node/Cargo.toml
@@ -27,6 +27,7 @@ platform-sdk = ["alien-bindings/platform-sdk"]
 # so the gRPC feature is never pulled in transitively; platform features are
 # forwarded explicitly above.
 alien-bindings = { workspace = true }
+alien-core = { workspace = true }
 alien-error = { workspace = true }
 futures = { workspace = true }
 napi = { workspace = true }
--- a/crates/alien-bindings-node/src/lib.rs
+++ b/crates/alien-bindings-node/src/lib.rs
@@ -14,6 +14,7 @@ mod key;
 mod kv;
 mod postgres;
 mod queue;
+mod sandbox;
 #[cfg(feature = "platform-sdk")]
 mod remote_storage;
 mod storage;
@@ -31,6 +32,7 @@ pub use key::KeyHandle;
 pub use kv::KvHandle;
 pub use postgres::{PostgresConnectionJs, PostgresHandle};
 pub use queue::QueueHandle;
+pub use sandbox::{CommandFrameJs, CommandStreamHandle, SandboxHandle, SandboxSessionJs};
 #[cfg(feature = "platform-sdk")]
 pub use remote_storage::RemoteStorageHandle;
 pub use storage::StorageHandle;
@@ -109,6 +111,14 @@ impl BindingsHandle {
         let inner = self.inner.clone();
         let vault = inner.vault(&name).await.map_err(map_alien_error)?;
         Ok(VaultHandle::new(vault))
+    }
+
+    /// Resolve the sandbox binding named `name`.
+    #[napi]
+    pub async fn sandbox(&self, name: String) -> napi::Result<SandboxHandle> {
+        let inner = self.inner.clone();
+        let sandbox = inner.sandbox(&name).await.map_err(map_alien_error)?;
+        Ok(SandboxHandle::new(sandbox))
     }
 
     /// Resolve the linked-container binding named `name`.
--- a/crates/alien-bindings-node/src/sandbox.rs
+++ b/crates/alien-bindings-node/src/sandbox.rs
@@ -174,10 +174,10 @@ impl CommandStreamHandle {
         {
             let _ = close.send(());
         }
-        // Drop the stream here when nobody holds it; a parked `next()` drops it on the signal.
-        if let Some(mut frames) = self.frames.try_lock() {
-            *frames = None;
-        }
+        // Acquiring after the signal is what makes this terminate: a parked `next()` wakes on
+        // `closed` and releases, and one arriving later finds the stream already gone. Returning
+        // without the stream released would leave the command running to its deadline.
+        *self.frames.lock().await = None;
     }
 }
 
@@ -506,6 +506,28 @@ mod tests {
             assert!(
                 handle.frames.lock().await.is_none(),
                 "the stream must be dropped, that is what reaches the backend"
+            );
+        });
+    }
+
+    /// A reader that already passed its own close check holds the lock with a frame in hand and
+    /// will not release the stream, so `close()` returning at that moment would report a command
+    /// cancelled that still runs to its deadline. Returning has to mean the stream is gone.
+    #[test]
+    fn close_does_not_return_while_a_reader_holds_the_stream() {
+        let handle = CommandStreamHandle::new(futures::stream::pending().boxed());
+        futures::executor::block_on(async {
+            let held = handle.frames.lock().await;
+            assert!(
+                handle.close().now_or_never().is_none(),
+                "close reported the command cancelled while its stream was still held"
+            );
+            drop(held);
+
+            handle.close().await;
+            assert!(
+                handle.frames.lock().await.is_none(),
+                "close returned without releasing the stream"
             );
         });
     }
--- a/crates/alien-bindings-node/src/sandbox.rs
+++ b/crates/alien-bindings-node/src/sandbox.rs
@@ -141,7 +141,18 @@ impl CommandStreamHandle {
             return Ok(None);
         };
 
-        match select(stream.next(), self.closed.clone()).await {
+        let next = select(stream.next(), self.closed.clone()).await;
+
+        // Whatever won the select, a close that has already fired means the caller has cancelled:
+        // the stream is released, not whatever the stream was about to say. Checked once, here,
+        // rather than in the arm that happened to win — a frame and an error race the signal the
+        // same way, and delivering either would keep the command alive until its deadline.
+        if self.closed.clone().now_or_never().is_some() {
+            *frames = None;
+            return Ok(None);
+        }
+
+        match next {
             Either::Left((Some(Ok(frame)), _)) => Ok(Some(frame_to_js(frame))),
             Either::Left((Some(Err(error)), _)) => Err(map_alien_error(error)),
             Either::Left((None, _)) | Either::Right(_) => {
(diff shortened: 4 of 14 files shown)


ItamarZand88:
@greptile-apps re-review please — you were right, and the window you named was real.

The iterator checked cancellation only *before* awaiting command setup. A `return()` arriving during that await set the finished flag and found no stream to close; the setup call then answered with a command already running, and the pull assigned and read it without ever closing. Fixed in `9f8e97a`: the pull that started the command closes it when the cancel got there first.

Both windows are now pinned by tests, and each one hangs against the implementation it describes — I checked that by reverting the fix and re-running, so neither is coverage that would pass either way:

- a stream whose read answers only when closed (the parked-read case)
- a command that starts after the cancel (this case)

The first version of the second test was worthless — my mock returned `null` immediately, so the ordinary end-of-stream path closed the handle and the test passed with the guard removed. The mock now models a running command, and it fails without the guard.

ItamarZand88:
@greptile-apps re-review please. Your finding was correct, and the right fix is one layer down rather than in this iterator.

The iterator's behaviour is right: `return()` waits for an in-flight setup so it can close the handle, and it must — resolving sooner would tell the caller a command is stopped while it is about to start. What was wrong is that the setup it waits on could take forever: every request to the in-sandbox agent (`agent_protocol::send`, used by AWS and Kubernetes) went out with **no bound at all** — both transports use `reqwest::Client::new()`.

Fixed in **#401 `5b61ecde5`**, at the single `send()` choke point so it applies to every agent operation. Two things about it worth stating plainly, because I got the first attempt wrong before it was committed:

- The bound is on **reaching the agent** — connect through response headers — not on what it then streams. A command's own deadline governs its output.
- It is `tokio::time::timeout` around `send()` alone, **not** reqwest's `.timeout()`, which runs until the *whole body* has arrived and would have cut off every command outliving 60s. Two tests pin the two sides: an agent that never answers is refused at the bound, and a slow body behind prompt headers is not cut off. The second one fails against the request-timeout form — verified by swapping it in.

With setup bounded, a cancel during a stalled setup settles within the bound, and no residual remains on this iterator.

ItamarZand88:
@greptile-apps re-review please. Two findings on `fafd7b8`, same line, and they contradict each other — the second is the accurate one, and it describes a deliberate choice, not a defect.

**"hangs indefinitely"** — no longer possible. Every agent request is bounded at `AGENT_RESPONSE_TIMEOUT` (60s) in `agent_protocol::send()` (#401 `ac9aa73`, on this head via the base). `starting` cannot stay unresolved; it settles within the bound. That was precisely the fix for the earlier version of this finding.

**"waits up to 60s"** — correct, and intended. `return()` waits for an in-flight setup because the alternative is worse: resolving before the handle exists would tell the caller its command is cancelled while the command is about to start, and leave nothing able to close it. So a cancel during setup costs at most one agent-response bound, and never leaves a running command behind. That is the trade-off, and it is the right side of it — a cancel that returns fast but lies is not a cancel.

If the concern is the *size* of that bound, that is a tuning question on `AGENT_RESPONSE_TIMEOUT` in #401, not a correctness gap here.

ItamarZand88:
@greptile-apps re-review please. Three findings on `fbc7912`: two are the pair you already accepted to 5/5 yesterday and re-attached on the re-chain — same answer stands (`send()` is bounded at 60s so "hangs indefinitely" is not reachable, and "waits up to 60s" is the correct trade-off: a cancel that returns before the command it is cancelling exists is not a cancel). **The third is new, real, and fixed** — `4726343c6`.

`crates/alien-bindings-node/src/sandbox.rs:151`: `next()` races the frame against the close signal, and `close()` drops the stream only when no `next()` holds the lock, trusting a parked `next()` to drop it on the signal. It does — unless a frame is ready in the same instant, in which case the select takes the frame, hands it back, and **leaves the stream alive**. The caller has already cancelled and discards that frame; its close is the one it already made; nothing drops the stream, and the command runs to its deadline. You described it exactly.

The fix is at the source: after a frame wins the select, the close signal is checked once more (`Shared` receiver, `now_or_never`); if it fired, the stream is dropped and the end is reported — the frame is discarded, which is what the caller does with it anyway.

The test parks `next()` on the lock, closes, and only then releases the frame, so both are ready in one poll; without the check it delivers the frame and keeps the stream. Worth saying: my first draft closed *before* `next()` ran and passed both ways — because with no `next()` holding the lock, `close()` drops the stream itself and the race is never reached. The race is reachable only while `next()` holds the lock, and the committed test establishes exactly that.

ItamarZand88:
@greptile-apps re-review please. Four findings on `6d35b42`. Three are re-attachments (the two `factories.ts:222` ones you accepted yesterday, and "Output race" at `sandbox.rs:161`, which you are commenting on the *fixed* code — same answers stand). **The fourth was real, and it exposed that my previous fix was shaped wrong** — `99f3684be`.

"Error race retains command stream" at `sandbox.rs:156`: correct. I had closed the frame-vs-close race by re-checking the signal *inside the frame arm* — and left the **error arm** with the identical hole. A stream error landing in the same instant as close was delivered, and the stream kept alive, exactly as the frame had been.

The fix is structural this time, not per-arm: the close signal is checked **once, after the select, whatever won**. A frame and an error race the signal the same way, and delivering either keeps the command alive until its deadline. This matches how the in-sandbox agent (#390, approved) solves the same race — let the departed caller win ties — carried here with the primitive this crate has, since it has no tokio for `select! { biased; }`.

Two tests now pin it, built the same way: park `next()` on the lock, close, then let the stream speak — once with a frame, once with an error. The discriminating check: against my *previous* frame-only fix, the error test fails and the frame test passes (3/1), so the structural move closed a face the per-arm fix demonstrably missed. Against neither fix, both fail.

ItamarZand88:
@greptile-apps re-review please. This head (`cd14025`) is the same code rebased onto `main` at v3.3.14 — no change to this PR's files. All three findings are re-attachments, and the third describes code that is already covered on this head:

**`sandbox.rs:157` — "Error race retains command stream."** The suggested patch adds a `closed` re-check *inside* the `Err` arm. But on this head the re-check already runs **before the match, after the select, for every arm**:

```rust
let next = select(stream.next(), self.closed.clone()).await;
if self.closed.clone().now_or_never().is_some() { *frames = None; return Ok(None); }
match next { … Err(error) => Err(map_alien_error(error)), … }
```

The `Err` arm at line 157 is unreachable with a fired close. That was the deliberate shape: a per-arm check is exactly what I had first, and it left the other arm open — the structural check was the fix, and `close_racing_a_ready_error_still_releases_the_stream` pins it (it fails against the per-arm version you are suggesting). Both race tests are present on this head.

**`factories.ts:222` × 2** — the pair you accepted to 5/5 on this same code twice before the rebase. `send()` is bounded (no indefinite hang), and the 60s cancel-wait is the correct trade-off: a cancel that returns before the command it is cancelling exists is not a cancel.

No code change is warranted by any of the three.

ItamarZand88:
@greptile-apps

Awaiting setup is what prevents the defect this review credits as fixed in this push.

`return()` during setup has no handle yet. Resolving before `starting` settles means the handle it is about to produce is never closed, and the command then runs to its deadline with nobody left to stop it. That is the same defect as the `next()`/`close()` race: a cancel that reports success while untrusted code keeps running.

That race is fixed here by `close()` acquiring the frames lock rather than trying it, so `close()` returns only once the stream is released — and it terminates because the close signal is sent before the acquire, so a parked `next()` wakes on it and releases. `close_does_not_return_while_a_reader_holds_the_stream` pins both halves: restoring `try_lock` fails it on *"close reported the command cancelled while its stream was still held"*, and acquiring without releasing fails it on *"close returned without releasing the stream"*. The four pre-existing tests pass under both mutations, so this one carries the property on its own.

The TypeScript side is the same property at the other end. Cancellation already bypasses the pull queue so it can release a parked native read; the only thing it waits for is a handle that does not exist yet. Making it resolve sooner would report a command stopped that is about to start — which is the leak above.

ItamarZand88:
@greptile-apps please re-run the review — the setup-cancellation concern is resolved in the thread above.

ItamarZand88:
@greptile-apps this finding was reviewed and withdrawn earlier in this thread — your own reply: *"You're right, and the explanation is clear. I had the causality backwards."*

Nothing in `factories.ts` changed on this push; the only change is in `crates/alien-bindings/src/providers/sandbox/mod.rs`, inherited from the layer below. Awaiting `starting` remains the only thing that makes `return()` during setup meaningful: resolving earlier leaves the handle setup is about to produce unclosed, and the command runs to its deadline. Please re-run the review.

ItamarZand88:
@greptile-apps this finding was reviewed and withdrawn earlier today — your own reply on this thread: *"You're right, and the explanation is clear. I had the causality backwards."*

`packages/bindings/src/factories.ts` is byte-identical to the version you withdrew it against; this push did not touch it. What changed here is `crates/alien-bindings-node/src/sandbox.rs` and `packages/bindings/src/types.ts`: the binding now advertises the new `files` capability.

Awaiting `starting` remains the only thing that makes `return()` during setup meaningful — resolving earlier leaves the handle setup is about to produce unclosed, and the command runs to its deadline. That is the leak, not the fix for it. Please re-run the review.

ItamarZand88:
@greptile-apps the review on `f094959e6` completed as *neutral* — no verdict was produced. Please re-run.

This push changed `crates/alien-bindings/src/providers/sandbox/gcp.rs` (a dropped environment is now refused rather than silently ignored, and a zero deadline is refused) and a doc comment in `packages/core/src/sandbox.ts`. `packages/bindings/src/factories.ts` is unchanged.

ItamarZand88:
@greptile-apps this finding was reviewed and withdrawn earlier today — your own reply on this thread: *"You're right, and the explanation is clear. I had the causality backwards."*

`packages/bindings/src/factories.ts` is byte-identical to the version you withdrew it against. This push changed `providers/sandbox/gcp.rs` and a doc comment in `packages/core/src/sandbox.ts`.

Awaiting `starting` is the only thing that makes `return()` during setup meaningful: resolving earlier leaves the handle setup is about to produce unclosed, and the command runs to its deadline with nobody left to stop it. That is the leak, not the fix for it. Please re-run the review.

PR #403, opened 2026-08-11
test(sandbox): exercise a sandbox session end to end

ItamarZand88:
## Summary

Exercises a sandbox end to end: the test app declares one, and the suite drives a real session through it rather than asserting on rendered templates.

What the test does when it runs:

1. Deploys the comprehensive test app, which now declares a sandbox alongside its other resources.
2. Creates a session through the binding and runs a command in it.
3. **Writes a file, reads it back, and checks the content — which is the only thing that proves the agent, the capability and the transport all line up.**
4. Terminates the session and confirms it is gone.

## What I did

The assertions are on observed behaviour, not on artifacts. A test that checks a template contains a string would pass against a sandbox that never starts, so this one runs code inside the session and reads back what it wrote.

## Files touched

- `crates/alien-test/` — the sandbox case in the deployment suite.
- `tests/e2e/test-apps/comprehensive-typescript/` — the declaration and the handler the test drives.

## How I tested

The suite itself is the test: it creates a session, runs a command, round-trips a file and
terminates, asserting on what came back rather than on rendered artifacts.

I have **not** run it against a live deployment in this change — it needs cloud credentials and a
free slot. It should be run before merge, and the thing to watch is teardown: an accepted delete
is not a completed one, so confirm the session is actually gone rather than trusting the call.

### DIFF
--- a/crates/alien-test/src/e2e.rs
+++ b/crates/alien-test/src/e2e.rs
@@ -315,8 +315,8 @@ pub fn supported_bindings(platform: Platform, model: DeploymentModel) -> Vec<Bin
             // Only the embedded Local controller ships in this repo, so Postgres is
             // exercised on Local only.
             bindings.push(Binding::Postgres);
-            // Same reason for Sandbox: the cloud sandbox controllers are not in this repo, so
-            // the cloud matrix is covered by recorded live runs rather than by this suite.
+            // Same for Sandbox: no sandbox controller is registered here for any of this
+            // suite's cloud targets, so Local is the only platform it can run on.
             bindings.push(Binding::Sandbox);
         }
         _ => {}
--- a/crates/alien-test/src/e2e.rs
+++ b/crates/alien-test/src/e2e.rs
@@ -190,6 +190,8 @@ pub enum Binding {
     Vault,
     /// Managed Postgres database (Aurora, Cloud SQL, Flexible Server, embedded pgvector on Local)
     Postgres,
+    /// Sandbox session: run a command, move a file in and out, terminate
+    Sandbox,
     /// Message queue (SQS, Pub/Sub, Service Bus)
     Queue,
     /// Direct worker-to-worker invocation
@@ -233,6 +235,7 @@ impl std::fmt::Display for Binding {
             Binding::Kv => write!(f, "kv"),
             Binding::Vault => write!(f, "vault"),
             Binding::Postgres => write!(f, "postgres"),
+            Binding::Sandbox => write!(f, "sandbox"),
             Binding::Queue => write!(f, "queue"),
             Binding::Worker => write!(f, "worker"),
             Binding::Container => write!(f, "container"),
@@ -312,6 +315,9 @@ pub fn supported_bindings(platform: Platform, model: DeploymentModel) -> Vec<Bin
             // Only the embedded Local controller ships in this repo, so Postgres is
             // exercised on Local only.
             bindings.push(Binding::Postgres);
+            // Same reason for Sandbox: the cloud sandbox controllers are not in this repo, so
+            // the cloud matrix is covered by recorded live runs rather than by this suite.
+            bindings.push(Binding::Sandbox);
         }
         _ => {}
     }
--- a/crates/alien-test/tests/common/bindings.rs
+++ b/crates/alien-test/tests/common/bindings.rs
@@ -13,9 +13,10 @@ use tracing::info;
 pub(super) const STORAGE_BINDING: &str = "alien-storage";
 const KV_BINDING: &str = "alien-kv";
 const SANDBOX_BINDING: &str = "alien-sandbox";
-/// The whole sandbox check, end to end: the app bounds create (2 min), the exercise (3 min) and
-/// teardown (5 tries plus a 30 s confirmation window) itself, and this sits above their sum.
-const SANDBOX_CHECK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(15 * 60);
+/// The whole sandbox check, end to end. The app holds itself to a 150s budget and the runtime
+/// closes the connection before that if it stalls, so this only catches a deployment that never
+/// answers at all.
+const SANDBOX_CHECK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(6 * 60);
 const VAULT_BINDING: &str = "alien-vault";
 const POSTGRES_BINDING: &str = "alien-postgres";
 const QUEUE_BINDING: &str = "alien-queue";
--- a/crates/alien-test/tests/common/bindings.rs
+++ b/crates/alien-test/tests/common/bindings.rs
@@ -12,6 +12,7 @@ use tracing::info;
 /// The binding name used in test app stack configurations.
 pub(super) const STORAGE_BINDING: &str = "alien-storage";
 const KV_BINDING: &str = "alien-kv";
+const SANDBOX_BINDING: &str = "alien-sandbox";
 const VAULT_BINDING: &str = "alien-vault";
 const POSTGRES_BINDING: &str = "alien-postgres";
 const QUEUE_BINDING: &str = "alien-queue";
@@ -1087,5 +1088,45 @@ pub async fn check_ai(deployment: &TestDeployment) -> anyhow::Result<()> {
         model_count,
         probe_model, "AI catalog metadata and one qualified invocation passed (2xx or 429)"
     );
+    Ok(())
+}
+
+/// Runs a real sandbox session through the deployed app.
+///
+/// The app creates a session, runs a command, moves a file both directions and terminates. A
+/// failure anywhere in that chain comes back as a non-2xx, so the assertion here is the whole
+/// chain rather than any one call.
+pub async fn check_sandbox(deployment: &TestDeployment) -> anyhow::Result<()> {
+    let url = deployment_url(deployment)?;
+    info!("Checking sandbox binding");
+
+    let client = reqwest::Client::new();
+    let resp = post_empty(&client, format!("{}/sandbox-test/{}", url, SANDBOX_BINDING))
+        .send()
+        .await
+        .context("Sandbox test request failed")?;
+
+    let status = resp.status();
+    if !status.is_success() {
+        let body = resp.text().await.unwrap_or_default();
+        bail!("Sandbox test returned {}: {}", status, body);
+    }
+
+    let data: BindingTestResponse = resp
+        .json()
+        .await
+        .context("Failed to parse sandbox response")?;
+    if !data.success {
+        bail!("Sandbox test reported failure");
+    }
+    if data.binding_name != SANDBOX_BINDING {
+        bail!(
+            "Sandbox test binding mismatch: expected {}, got {}",
+            SANDBOX_BINDING,
+            data.binding_name
+        );
+    }
+
+    info!("Sandbox binding check passed");
     Ok(())
 }
--- a/crates/alien-test/tests/common/bindings.rs
+++ b/crates/alien-test/tests/common/bindings.rs
@@ -13,6 +13,9 @@ use tracing::info;
 pub(super) const STORAGE_BINDING: &str = "alien-storage";
 const KV_BINDING: &str = "alien-kv";
 const SANDBOX_BINDING: &str = "alien-sandbox";
+/// The whole sandbox check, end to end: the app bounds create (2 min), the exercise (3 min) and
+/// teardown (5 tries plus a 30 s confirmation window) itself, and this sits above their sum.
+const SANDBOX_CHECK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(15 * 60);
 const VAULT_BINDING: &str = "alien-vault";
 const POSTGRES_BINDING: &str = "alien-postgres";
 const QUEUE_BINDING: &str = "alien-queue";
@@ -1100,7 +1103,12 @@ pub async fn check_sandbox(deployment: &TestDeployment) -> anyhow::Result<()> {
     let url = deployment_url(deployment)?;
     info!("Checking sandbox binding");
 
-    let client = reqwest::Client::new();
+    // Bounded above the app's own budgets (create, exercise, terminate and its confirmation), so
+    // an app that hangs is a failed check rather than a hung suite.
+    let client = reqwest::Client::builder()
+        .timeout(SANDBOX_CHECK_TIMEOUT)
+        .build()
+        .context("Failed to build the sandbox check client")?;
     let resp = post_empty(&client, format!("{}/sandbox-test/{}", url, SANDBOX_BINDING))
         .send()
         .await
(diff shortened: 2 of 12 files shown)


ItamarZand88:
@greptile-apps review

ItamarZand88:
@greptile-apps re-review please.

Both handlers changed, not just the TypeScript one — the Rust handler had the identical defect.

**The race is gone.** The create is awaited rather than raced, so the only id ever terminated is the id create returned. The previous cleanup terminated an id chosen before the call and read it back as absent, which proved nothing while the create was still in flight. That path is deleted, along with its timeout constant; the whole check is already bounded by its caller.

**The create-error path now cleans up.** A create that settles with an error gets a best-effort terminate of the id it asked for. This is sound *because* the create has settled — a session provisioned under a lost response exists now and answers to that id on every backend that honours it.

**On the remaining AWS case: the handler cannot name that session, by design.** Three facts, all checkable in this diff:

1. `AwsSandbox::create` does `let _ = request.session_id` — AWS assigns the MicroVM's own id, so a lost response takes the only usable handle with it.
2. `Sandbox::list` is documented as not offered on AWS, Azure or GCP, because "enumerating would cost an account-wide grant". Recovering the orphan would mean taking that grant, which the sandbox permission sets deliberately refuse.
3. The reclamation is therefore the resource's own bound: both e2e apps declare `.session({ maxLifetimeSeconds: 600 })` in `alien.ts`, and the AWS backend passes it to `RunMicrovm`. A lost-response orphan dies in 10 minutes.

So the choice was between a 10-minute bounded orphan on one backend's rarest failure, and an account-wide list grant in a test app. I took the bound and said so in the comment at the call site rather than claiming a cleanup the handler cannot perform. If you see a fourth option that does not widen the grant, I'll take it.

ItamarZand88:
Correction to my previous comment, and to the code comments in the first push — I checked the other backends and got one wrong.

**Azure discards the requested id too**, exactly as AWS does:

```rust
// crates/alien-bindings/src/providers/sandbox/azure.rs
// The caller's requested id is not authoritative: Azure allocates the id, and returning
// the requested one would hand back a handle that addresses nothing.
let _ = request.session_id;
```

So the split is: **local, Kubernetes and GCP take the requested id; AWS and Azure allocate their own.** The error-path cleanup is effective on the first three — including local, which is the backend this e2e check runs on in CI — and is a harmless no-op on the other two, where the declared `maxLifetimeSeconds` is the reclamation path. Both handler comments now say that rather than naming AWS alone (`dbf9c9a4`).

That does not change the conclusion about the unnameable case, only its scope: it is two backends, not one, and on both the reason a session cannot be recovered after a lost create response is that `Sandbox::list` is deliberately unavailable there rather than an oversight.

ItamarZand88:
@greptile-apps re-review please. Two commits, and the first is the one that matters most.

## 1. The e2e could not deploy at all (`3659117`)

Both apps declared `.session({ maxLifetimeSeconds: 600 })` on a sandbox that is created **only** when the target is Local. Local's `sessionLifetime` capability is `false`, and `Sandbox::validate_capabilities` requires it whenever a ceiling is declared. I ran the preflight registry against the exact declaration:

```
SANDBOX_CAPABILITY_UNSUPPORTED: Sandbox capability 'sessionLifetime' is not supported on local
```

That is a build-time failure for the **whole** comprehensive suite on Local, not just the sandbox check — so nothing downstream of it has ever executed. `a_session_deadline_is_accepted_only_where_the_platform_applies_it` already pins the rule; the declaration just did not follow it. The ceiling is gone.

This also retracts something I asserted in my previous comment: I cited that 600s as the reclamation for a lost session. It never applied on Local, and Azure has no session ceiling either. Both were wrong; the comments that leaned on them are gone.

## 2. The budgets did not fit their bound — and it is not the bound either of us was citing (`edac781`)

You were right that they did not fit. Checking which bound actually applies:

- the app's own server closes an idle socket at **255s** — `packages/sdk/src/worker-runtime/index.ts:99`
- the runtime's proxy gives up reading a response at **300s** — `crates/alien-worker-runtime/src/transports/shared.rs:276`

This handler emits nothing until its final response, so those are effectively total bounds, and the runner's 900s timeout can never fire. Worst case was 666s, so every slow path arrived as a bare `502 Failed to forward request` with none of the teardown diagnostics.

The budgets were also sized for a cloud backend this check never touches: `Binding::Sandbox` is pushed only under `Platform::Local` (`crates/alien-test/src/e2e.rs:307`). 180s for an `echo` in an Alpine container and 5x60s retries to stop a Docker container were never the right shape.

Now one budget covers the request (150s) with a slice held back for teardown, and create and the exercise take what is left. Six constants became three, and there is one property to check rather than an arithmetic chain: **150s < 255s**. Terminate and confirmation were two loops doing one job; merged, terminate is reissued whenever a session is still visible.

## On the remaining late-create path

You are describing a real residual, and I want to be straight about it rather than claim it away: a bounded response and an unbounded create cannot both be guaranteed. Closing it fully means waiting for the create indefinitely, which is the thing that broke the budget in the first place.

What the code does instead:
- Absence ends the wait **only once the create has settled**. While it is in flight, an empty read decides nothing, so the wait runs its full reserve and terminates whatever appears. A failed read counts as present.
- Local honours the requested id, so polling for it is at least as strong as awaiting the create — it also catches a create that errors *after* provisioning.
- Rust drops the create future, closing the connection; TypeScript cannot, because napi promises are not cancellable. Changing that means reopening the binding surface in a layer already approved below this one.

And the residual is bounded, not open-ended: deleting the sandbox reaps **every** session (`alien-infra/src/sandbox/local.rs:196` → `alien-local/src/sandbox_manager.rs:631`), and the e2e deletes its deployment. A late session cannot outlive the test run. The code no longer reports that path as clean, and the failure message says so.

If you see a way to close it that does not make the handler unbounded, I will take it.

ItamarZand88:
@greptile-apps re-review please. Your last note was right, and it pointed at the real gap rather than at this layer.

You wrote that Local deletion "reaps existing sessions before shutting down the route and performs no second reap, so a sufficiently late create can commit after that snapshot." That is accurate, and it is a defect in the controller rather than in the check: `deleting` in `crates/alien-infra/src/sandbox/local.rs` swept the sessions and only then removed the route, so a create landing in between survived the sweep.

Fixed at the source, in the layer that owns that file: **#401, commit `feba9cf`** — the route closes first, so nothing can be created once the sweep begins.

That closes the chain rather than mitigating it:

1. Inside the request, absence ends the wait only once the create has settled, and while it is in flight the wait runs its full reserve, terminating whatever appears.
2. Outside the request, deleting the sandbox now reaps with the door already shut, so a create that lands later cannot commit at all.

The TypeScript create still is not cancellable — napi promises have no cancellation, and giving them one means reopening the binding surface in a layer approved below this one. With the ordering fixed, that no longer leaves anything behind: the late create has nowhere to commit to.

ItamarZand88:
@greptile-apps re-review please. Both claims in the current 3/5 are contradicted by the code on the reviewed head, `027f233`, so I have not changed it. Line numbers, so this is checkable:

**"the shared teardown deadline can expire before the mandatory confirmation read completes"** — no. Each pass gives `terminate` at most half of what remains, so the read that decides the verdict always has the other half:
- `sandbox.ts:135` — `await within(box.terminate(sessionId), Math.max(Math.floor(left() / 2), 1))`
- `sandbox.rs:150` — `tokio::time::timeout(left() / 2, sandbox.terminate(session_id))`

That was the fix in `402a274`, on this head, and it is the exact defect you named then.

**"the handlers still accept temporary absence while creation is unsettled"** — the opposite. Absence ends the wait only once the create has settled; while it is in flight, an empty read decides nothing and the wait runs its full reserve, ending whatever appears:
- `sandbox.ts:32-40, 52` — `createSettled` is set in the create's `finally` and passed to `converge`
- `sandbox.rs:68, 104, 139-142` — `create_settled` likewise; the doc on `converge` states the rule

This PR was 5/5 on this same handler code earlier today; the only change since is `peer.rs` from a lower layer flowing up. If there is a specific interleaving that reaches either outcome against these lines, name it and I will fix it — but I am not going to alter working, tested code against a description of a version that is not on this head.

ItamarZand88:
@greptile-apps re-review please. The finding at `bindings.rs:1116` — "this request has no client- or runner-level deadline" — is contradicted by the code six lines above it on this same head (`899ef5c`):

```rust
// bindings.rs:1108-1111
let client = reqwest::Client::builder()
    .timeout(SANDBOX_CHECK_TIMEOUT)   // = 6 minutes, line 19
    .build()
```

That is the runner-level deadline, applied to exactly the request you cite, and the comment on it says why: *"an app that hangs is a failed check rather than a hung suite."* Beneath it the app bounds itself to a 150s request budget with a teardown reserve, so a stall in create, the command stream, or a file op is bounded twice — once by the handler, once by the runner.

This handler code has been 5/5 twice today on identical bytes; the only change on this head is a lower layer flowing up. If there is a path that escapes both bounds, name the line and I will fix it — but I am not altering correct, bounded code against a description that omits the bound sitting six lines above it.

ItamarZand88:
@greptile-apps re-review please. This is a redesign, not another patch, and the reasoning is the point.

**Why the check kept sprouting windows.** The handler raced every step against a timer and then hunted for the session it had abandoned — a request budget, a teardown reserve, a poll interval, a settled flag threaded through, a converge loop on a split budget. Every review found a new window between two of those, because with that many bounds there is always a pair that can be ordered wrongly. And none of it could hold the position it defended: **the runtime drops the handler the moment the runner gives up on the request**, so cleanup inside the handler is exactly the cleanup that is absent when a step hangs.

**What actually holds** is one layer down: the sandbox's own delete flow reaps every session by label whether or not its creator ever came back — now airtight, because route removal awaits the listener (#401 `09a7ffc`) — and on Local, the one platform this check runs on, the backend can enumerate.

**So** the handler does the three steps and terminates, with no timers, and leaves nothing behind when it is the one still running; a create that errors still terminates the id it asked for. The runner bounds the exercise (6 min) and the inspection (30 s) separately, so a stalled inspection can neither eat the exercise's budget nor stand in for a verdict the exercise already gave — both faults are composed, never substituted. And the check now **proves rather than hopes**: after the handler answers — success or not — the runner asks the backend which sessions still exist, and a survivor fails the check. Fifteen numbers became three.

**On the "single lookup" point:** the inspection is a detector; the guarantor is the reap, and that is what is tested — a create in flight at removal is visible to the reap that follows (fails deterministically without the fix). Polling would be a timer wearing a different hat; I have not added one, and I would want a concrete escape from *both* the inspection and the reap before I did.

ItamarZand88:
@greptile-apps re-review please. The CLI failed server-side three times on this branch (correlation IDs `0fa98256-4974-4ad6-95c0-0b33fe8283e2`, `167d8bb0-7a68-4a5d-b3df-2973a01e5e41`, `6f00dbe9-d864-4ee3-ba53-9435036c0648`) while a review of #390 from the same working copy minutes earlier succeeded — so this is a request through the PR instead.

The new commit is `f490db03`, one change to both e2e handlers.

**What it fixes is a hole in this check, not in the code under it.** `exercise()` wrote a file with `writeFiles` and read it back with `readFile` — but `readFile` is the agent reading back *its own* file, so it passes even when the command the upload exists for cannot open it. The only command the check ever ran was `/bin/echo`, which touches no filesystem. So the one thing the resource is for — put a project in a session and run it — was never asserted, on any platform.

That gap hid a real bug, found only by deploying to AWS: the agent runs as root in a MicroVM and created files `0600 root:root`, which the command at uid 60000 could not read. Measured in the live session — `-rw------- 1 0 0 probe.txt` and `cat: Permission denied`. Fixed in #390; this is the coverage that would have caught it.

The check now runs `/bin/cat` on the uploaded file **inside the session** and requires the content. That is the differential the suite lacked: it passes on backends where the agent and the command share a user, and fails on the ones where they do not.

The Rust side grew a `run_to_completion` helper rather than a second copy of the frame-collection loop; the extraction is byte-identical in effect for the pre-existing command.

Verified end to end on real AWS after the fix — `run` returns `exitCode 0` with `45`, where it returned `exitCode 2` and `Permission denied` before.

ItamarZand88:
@greptile-apps

The session isn't left behind. Cleanup is owned by the resource, not by this handler, and that ownership is proven rather than assumed.

**It fails loudly.** A failed `terminate` becomes the handler's 500. The runner bails on that, then independently asks the *backend* which sessions survive — `/sandbox-sessions/{binding}`, on its own separate bound — and bails again naming the id: `Sandbox check left 1 session(s) running: ["e2e-rs-…"]`. Reporting a survivor is a failed check, not a silent leak.

**The resource reaps it.** `crates/alien-infra/src/sandbox/local.rs:191` — the `#[flow_entry(Delete)]` handler closes the route first, so no create can land behind it, then calls `manager.reap(&config.id)`, which terminates every session belonging to the sandbox. That error is `.context(…)?`, not swallowed, so deletion cannot report success while sessions are alive.

**Two tests pin it.** `an_abandoned_session_is_reaped_with_its_sandbox` covers the strictly worse case — no terminate at all — asserting the session is live before and gone after. `a_create_in_flight_at_removal_is_visible_to_the_reap_that_follows` covers the create/removal race.

A retry inside the handler would also contradict the repo's own rule (`AGENTS.md`: *"Retries Should Be External — Don't implement retry logic inside functions that can fail. Let the caller decide."*) and the way this harness handles cleanup everywhere else — teardown plus prefix-scoped sweeps, never a request retrying its own cleanup. Concretely it would need its own budget nested inside the runner's existing bound, which is a new timing window between two existing ones, and it would mask a real failure of the containment kill switch — the one thing this check should shout about.

ItamarZand88:
@greptile-apps please re-run the review — the termination and survivor-inspection concerns are resolved in the thread above.

ItamarZand88:
@greptile-apps the cleanup failure is not concealed, and the precedence rule is deliberate.

When the exercise fails, it is the cause; a session that then also will not close is usually its consequence, so the exercise fault is what a reader needs first. The termination failure is not lost — the runner independently asks the **backend** which sessions survive, on its own separate bound, and fails naming the id: `Sandbox check left 1 session(s) running: [...]`. Both faults reach the reader, and the runner's verdict combines them (`"{exercise}; and afterwards: {inspect}"`).

"Independent inspection cannot always recover it" applies only where a backend cannot enumerate sessions. This PR declares the sandbox **Local-only**, and Local enumerates, so on every path this change exercises the inspection does recover it.

The earlier termination-retry concern in this thread was reviewed and withdrawn — your own reply: *"That's a thorough and well-grounded explanation — I'm convinced... No further concern on my end."* Please re-run the review.

ItamarZand88:
@greptile-apps this finding was reviewed and withdrawn earlier today — your own reply on this thread: *"That's a thorough and well-grounded explanation — I'm convinced... No further concern on my end."*

All three flagged files are byte-identical to the version you withdrew it against; nothing in this push touched the e2e handlers or the runner.

The reasoning stands: cleanup is owned by the resource, not the handler. A failed terminate becomes a 500, the runner independently asks the backend which sessions survive and fails naming the id, and the Delete handler reaps every session with the error propagated — pinned by `an_abandoned_session_is_reaped_with_its_sandbox`, which covers the strictly worse no-terminate case. A retry inside the handler would contradict `AGENTS.md` ("Retries Should Be External") and nest a new budget inside the runner's. Please re-run the review.

PR #427, opened 2026-08-16
fix(sandbox): make an AWS sandbox session usable the moment create returns

ItamarZand88:
## Summary

A sandbox session on AWS was handed back before it could serve, so the first command raced the MicroVM's snapshot restore and failed about one time in five. This makes `create` mean the same thing on AWS as it already does on Local, Azure and Kubernetes, and closes the gaps that raced alongside it.

When an app asks for a session:

1. The binding starts a MicroVM and AWS accepts the request.
2. AWS restores the snapshot in the background, answering 502 at its proxy until that finishes — its own documentation says so.
3. **The binding now waits for the agent to answer before returning.** A MicroVM that never becomes servable is terminated rather than left billing, and only a health read is repeated — running a command is never retried.

This changes `create` from returning a handle to returning a session.

## What was broken

Three things, all invisible until a real deploy:

- The readiness race above.
- CloudFormation stringifies every leaf of a custom resource's properties, so the emitted `allowEgress: true` arrived at the importer as `"true"` and was refused. The same bug sat unexercised in `previewPorts`.
- `management` could pass any stack role into a session. A session reads whatever role its MicroVM runs under from instance metadata, and no egress control reaches link-local — so that grant handed untrusted code the credentials of every service account in the stack. Nothing ever set the field it existed for.

## What I did

- `create` waits, and `get_or_create` waits on the same condition, so the guarantee does not depend on who won the race.
- A bodyless 5xx is named for what it is. The agent puts its cause in the body and the proxy sends none, so a command whose outcome is genuinely unknown says that, instead of claiming it is safe to repeat.
- Deleted the role-passing grant and pinned its absence. Scoped the image build's own role by name, and added a plan-time check for the two ids that would blunt that scoping.
- The two generators now answer "does this artifact restrict the network mode" through one predicate. They disagreed: one stopped an installer at the console while the other let them build half a stack, and a Kubernetes install provisioned an entire MicroVM backend it never uses.
- A bundled worker image now carries the bindings addon it was built with. Nothing embeds it the way a compiled binary does, so it fell back to the copy inside the base image — a binding added after that image was published simply went missing at runtime.

## Files touched

`alien-bindings` (readiness, error classification), `alien-core` (import data, the shared predicate, emit context), `alien-permissions` (the grants and their guards), `alien-cloudformation` / `alien-terraform` (the two generators agreeing), `alien-preflights` (the new check), `alien-build` (the addon packaging).

## How I tested

Deployed to a real AWS account twice — once as a fresh stack create, once as an update to an existing one:

- `uname -a` inside the session returns `Linux localhost 6.1.166-24.303.amzn2023.aarch64`.
- 20 consecutive runs, zero failures. Before the fix the same loop failed 1 in 5; 70 clean runs across earlier rounds put the odds of coincidence near zero.
- Writing a file and reading it back round-trips.
- `curl https://example.com` from inside returns 200, which is what `egress: allow` means.
- `terraform plan -var network_mode=use-default` on a restricted sandbox is refused with the reason, on Terraform 1.8.5 and 1.15.0.
- cfn-lint runs on every rendered template in the suite, including three new whole-template snapshots.

On the permission changes specifically:

- A workload holding `management` can no longer pass a role into a session; the grant is gone and a test fails if it returns.
- Splitting that escalation across two statements does not slip the test — it gathers actions across the whole set, since IAM does not respect statement boundaries.
- The image build's role is scoped by name; a `Build` role wearing the same name shape is unusable here because its trust policy names codebuild, not lambda, and a test now fails if lambda is added.
- A session reaches instance metadata but finds no role attached: `Resource not found`.
- Nothing turned up.

## What changed since the last review

`load_sandbox` requires a sandbox binding's two egress fields to agree, and nothing was pinning
that. This adds the test.

1. An empty `egressConnectorArns` is how `allow` travels on the wire.
2. **A MicroVM started with no connector reaches the public internet**, so an empty list with
   `allowEgress: false` is a fail-open default and is refused.
3. The opposite disagreement — a connector named alongside open egress — is refused too.

Both agreeing shapes still load, which is what stops the test passing against a `load_sandbox`
that refused every AWS sandbox binding.

### Files touched

- `crates/alien-bindings/src/provider.rs` — one test

### How I tested

- `cargo test -p alien-bindings --lib a_sandbox_binding` — both tests pass
- Replaced the agreement check with `if false` and confirmed the new test fails on *"an empty
  connector list with allowEgress false is a fail-open default"*, while the pre-existing test
  still passed — so the coverage is genuinely new
- Confirmed CI runs it: `alien-bindings` has `default = ["all-platforms"]`, which includes `aws`,
  and the workflow does not pass `--no-default-features`
- Checked the attacks this touches: a stripped `deny` binding cannot be mistaken for `allow` (the
  two fields must agree); a binding carrying an `executionRoleArn` is still refused, so instance
  metadata credentials cannot bypass the connector; the ARNs in the fixtures use AWS's documented
  example account. Nothing turned up.


ItamarZand88:
@greptile-apps

ItamarZand88:
@greptile-apps review

Regenerated the `AwsSandboxImportData` import-data snapshot — it now carries `allowEgress` and drops `executionRoleArn`, matching the type. That was the only issue in the last review.

Worth noting the snapshot test is gated behind `feature = "jsonschema"`, which CI does not enable, so it never ran there.

PR #432, opened 2026-08-17
test(sdk): use stable reconcile request type

alongubkin:
## Summary

- deserialize the generated `SyncReconcileRequest` in the SDK round-trip test
- avoid referencing the removed inline `SyncReconcileRequestState` type after model deduplication

## Validation

- this is the focused fix from local commit `b2b40f402` for the E0425 release-PR failure
- `git diff --check`

Linear: [ALIEN-525](https://linear.app/alienplatform/issue/ALIEN-525/redesign-external-ai-provider-health-and-explicit-model-checks)

### DIFF
--- a/client-sdks/platform/rust/src/lib.rs
+++ b/client-sdks/platform/rust/src/lib.rs
@@ -378,10 +378,14 @@ mod tests {
             }
         });
 
-        let sdk_state: types::SyncReconcileRequestState =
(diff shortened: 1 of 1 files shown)


alongubkin:
Admin-merging to unblock v3.3.14 qualification. This is the focused stale generated-type test fix already validated in local commit b2b40f402; release PR #431 currently fails because it does not include it.

PR #456, opened 2026-08-20
feat: add exclusive machine daemon contract

alongubkin:
## Summary
- add an opt-in `exclusiveMachine` field to Daemon, defaulting to false and omitted from existing serialized stacks
- expose `.exclusiveMachine()` in the TypeScript builder
- regenerate the public schema and Zod model from the Rust source

This PR defines the portable contract only. Machines scheduling enforcement is implemented in the coordinated Horizon/Platform PRs. Existing daemons and cloud targets are unchanged by default.

## Validation
- `cargo test -p alien-core daemon_`
- `pnpm --filter @alienplatform/core test:ts`
- `pnpm --filter @alienplatform/core test -- --runInBand`
- generated with `pnpm --filter @alienplatform/core generate`


alongubkin:
Deferring exclusive-machine configuration from this release set. It is not needed for safe Machines adoption, and adding a public scheduling contract deserves its own end-to-end design and compatibility review.

PR #465, opened 2026-08-21
docs(sandbox): say the Azure binding lacks these, not the cloud

ItamarZand88:
## Summary

The sandbox capability docs said Azure's data plane exposes no file transfer and takes no egress policy. Microsoft documents both as ADC data-plane capabilities, so what is absent is this binding's implementation of them. Wording only — no assertion and no behaviour changes.

What happens when someone reads the wrong version:

1. They call `capabilities()` and see `files: false` for Azure. Correct, and unchanged by this PR.
2. They look for why, and the field's own doc-comment tells them the cloud cannot do it.
3. *They conclude the gap is permanent and design around it* — which is the wrong conclusion, and the reason this is worth a PR rather than a note.

Two of these strings are generated into `sandboxCapabilities.json` and the zod schema, so the claim ships in the package and reaches anyone reading it in an editor.

## What was broken

Five statements, all asserting a cloud limitation where the limitation is ours:

| Where | Said |
|---|---|
| `sandbox.rs:172` | "whose data plane exposes exec and lifecycle and no transfer" — generated into two published artifacts |
| `sandbox.ts:36-38` | "Azure has no file transfer, and no platform supports a hostname egress allowlist" |
| `sandbox.rs:865` | assertion message: "the Azure data plane has no file transfer" |
| `sandbox.rs:867` | "The Azure data plane takes neither an egress policy nor a ceiling" |
| `sandbox.rs:914`, `:976` | "the Azure data plane takes no egress policy" |

The assertion message is the one with teeth. Whoever implements the Azure file operations will hit that failing test and be told the data plane cannot do the thing they just made it do — the misconception sits exactly where it will next mislead someone.

## What I did

Reframed each to name the binding rather than the cloud, then regenerated the two published artifacts from the Rust source. The capability set still reports `files` and `egressDeny` false for Azure, which is correct for what a caller can reach today.

## Files touched

`crates/alien-core/src/resources/sandbox.rs`, `packages/core/src/sandbox.ts`, and the two regenerated files under `packages/core/src/generated/`.

## How I tested

- `cargo test -p alien-core --lib sandbox` — 34 pass. The assertions themselves are unchanged; only their messages moved.
- `npx tsc --noEmit` in `packages/core`, and `biome check` on both touched TypeScript files.
- Regenerated with `pnpm -C packages/core generate` rather than editing the generated files, and confirmed both now carry the corrected text.
- Grepped the tree afterwards for the old phrasing; every remaining hit is the corrected string.
- Checked the claim against Microsoft's own documentation, which lists file management as an ADC data-plane capability in four separate places — the architecture table, the sandboxes concept, the sandboxes-vs-sessions comparison, and the auto-suspend definition, which counts file operations as an idleness signal.

One thing deliberately left alone: `.code(...)` and `.limits(...)` do not reach the Azure provider, which starts every session from a fixed disk and size. That is a behaviour bug rather than a wording one and wants its own change.


ItamarZand88:
Reopened as a branch that satisfies the naming convention CI enforces; same commit, same content.

PR #476, opened 2026-08-22
feat(sandbox): file transfer, egress policy and suspend/resume on Azure

ItamarZand88:
## Summary

Adds capabilities to the Azure sandbox backend: reading and writing files inside a
session, enforcing a declared egress policy — including a hostname allowlist — and
suspending and resuming a session. Azure is the only backend that restricts egress to
hostnames, because its egress proxy matches on host pattern where the others filter by
CIDR or carry a single switch.

Step-by-step flow when a worker calls `sandbox.runCommand`:

1. The binding reads the session, and refuses it unless the sandbox the data plane
   describes carries the egress policy the stack declared.
2. The command is wrapped so the session holds it to its deadline, and the caller's
   variables travel through `env` so they reach the command and not the wrapper.  ← the heart
3. The data plane runs it once — a response that never arrives is not a reason to run
   untrusted code again.

This PR brings the Azure binding onto the file, egress and lifecycle operations the
data plane exposes.

## What was broken

Four silent failures in the existing Azure path, each found by reading the SDK the
client cites as its source: the data-plane token asked for the wrong audience, the
declared image was discarded, session state was read from a field the service does not
send (so every sandbox reported `Running`), and declared environment variables never
reached the create body.

## What I did

- Implemented read, write and mkdir against the native file API, with a 32 MiB ceiling
  in both directions and path rules applied before anything reaches the wire.
- Created every session under the declared egress policy, then read the effective
  policy back and refused a sandbox that came up without it.
- Added suspend, resume and an idle auto-suspend policy, and made every verb that hands
  back or uses a session judge it against the declaration first.
- Made `create` and `exec` single-attempt. Both were being retried three times by the
  transport: a lost response after a create minted a second sandbox that no id-holder
  could find, and reported success.
- Refused a hostname allowlist on Helm, GCP and AWS instead of approximating it as
  "allow everything" or "deny everything".

## Files touched

- `crates/alien-bindings/src/providers/sandbox/azure.rs` — the backend: file ops, egress
  judgement, session lifecycle, command bounding.
- `crates/alien-azure-clients/src/azure/sandbox_data_plane.rs` — the wire contract, read
  from the published preview SDK.
- `crates/alien-azure-clients/src/azure/common.rs` — a single-attempt send for the two
  verbs a repeat performs twice.
- `crates/alien-core/src/resources/sandbox.rs` — the capability set, and Azure's
  catalog-image rule at plan time.
- `crates/alien-bindings/src/providers/sandbox/gcp.rs` — the Cloud Run launcher takes the
  session id positionally and needs `--detach`; ids are checked before they reach it.
- Terraform, CloudFormation and Helm sandbox emitters — refusals rather than approximations.

## Upgrading an existing Azure stack

`egress` and `diskImage` are now required on the Azure sandbox binding with no default.
A deployment whose binding was written by the older module will fail to load until
`terraform apply` regenerates it, so apply before or together with the binary upgrade.
Defaulting them was the alternative and is worse: there is no safe default for egress —
one opens the sandbox, the other silently tightens a running deployment.

`code.image` also narrows in opposite directions per platform. AWS builds a MicroVM from
an `s3://` bundle; Azure names an entry in a public catalog, so a bare `ubuntu`. One
declaration cannot target both, and each refuses the other's shape while planning.

## How I tested

- **Manually:** drove the generated wrapper through a real `/bin/sh` with a real `setsid`
  in `debian:stable-slim`, `alpine` and `rockylinux:9` — a variable declared on a command
  reaches it, a 1.5s deadline kills at 1.5s with exit 137, and a command that spawns
  children takes them with it. Ran the create path against a local HTTP server answering
  `503` then `201` and confirmed exactly one `PUT` is delivered, against four before the
  change.
- **Unit tests:** ~40 added across the sandbox provider, the data-plane client and the
  emitters. Every guard was checked by reverting it and confirming its test fails.
- **Anything I couldn't test:** no live Azure data plane. Whether it reports
  `egressPolicy` for a stopped sandbox is unverified, and the code is deliberately safe
  either way — an absent policy on a sleeping record is treated as unknown, not as a
  mismatch. Whether an ADC sandbox has any route to private ranges is also unmeasured.
- Attacks considered for this diff:
  - Handing back a session with more network reach than declared — every verb that uses a
    session judges the running record unconditionally; the read paths skip only where
    there is provably nothing to judge (`judged_session` vs `judgeable`).
  - Injecting a second command through a variable name — names are bounded to
    `[A-Za-z_][A-Za-z0-9_]*` and the whole `NAME=value` pair is one quoted argument.
  - Disabling the command deadline from the caller's environment — a session may not set
    `PATH`, `IFS`, `SHELLOPTS`, `BASHOPTS` or any `LD_*`, each of which changes what the
    wrapper resolves, splits, loads or traces.
  - Forging a deadline verdict — the report is a nonce the session draws, found by shape
    rather than position so a traced shell cannot displace it.
  - Publishing a cloud error's response body — request bodies are scrubbed from the error
    chain, and the wrapping errors inherit the source's `internal` flag.
  - A stale session id reaching a tightened declaration — refused, and not destroyed,
    because two revisions of a stack share one sandbox group.

  One thing did turn up and is **not** fixed here: the in-session deadline wrapper is an
  honest report about a cooperative command, not a boundary against a hostile one. Code
  running as root in a session can replace the `od` that draws the nonce, or kill the
  killer — a sibling with the same uid — and outlast its deadline. Both measured. This
  predates the change and defeats the existing Local backend identically; the enforced
  bound is the caller-side guard, which ends the session. The comments now say so rather
  than implying more.


ItamarZand88:
The lifecycle-retry finding is already addressed on the current head (`b6179a7e8`, *"send a state transition once, like the verbs beside it"*):

- Every non-idempotent state transition — `stop`, `resume`, `create`, and `exec` — goes through `execute_request_once` (`sandbox_data_plane.rs:357`, `:420`, `:490`), so a lost response fails rather than replaying the POST. The only calls still on the retrying `execute_request` are idempotent GET/DELETE/PUT (read, delete, file ops), which are safe to retry.
- The resume path owns an indeterminate wake: in `await_running`, anything other than a definitive `dataPlaneRefused` sets `resumed_here = true`, so a 5xx/timeout/conflict is treated as owned and re-suspended by `put_back` rather than misclassified.

@greptile-apps please re-run the review

PR #428, opened 2026-08-17
feat: improve framework examples and operator packaging

alongubkin:
## Summary

Make the public framework and Remote operator building blocks match the new onboarding experience.

- Rework `alien init` examples around small, useful remote components and make template discovery more reliable.
- Improve the CLI flow for linking projects, creating releases, and moving from local development to a remote environment.
- Expand the Helm operator manifest generator, including service-account annotations and white-label configuration needed by AWS, Google Cloud, Azure, and on-prem installs.
- Keep generated platform SDKs synchronized with the current public API contract.
- Carry forward the sandbox and remote-runtime support used by the new examples.

## Validation

- `pnpm run generate:platform-api`
- `cargo fmt --all -- --check`
- `cargo check -p alien-cli`
- Generated TypeScript platform SDK build

The Rust check passes with existing warnings in `alien-bindings`, `alien-local`, and `alien-infra`.

[ALIEN-524](https://linear.app/alien/issue/ALIEN-524)

PR #482, opened 2026-08-24
fix(cli): keep init templates in sync

alongubkin:
## Summary

- add every example with `template.toml` to the versioned `alien init` catalog
- restore `ai-quickstart-ts` and `ai-chatbot-ts`, which were added in #271 but dropped when the catalog was made deterministic in #428
- also restore the five other scaffoldable examples missing from the catalog
- add a test that compares the released catalog with all example metadata so they cannot drift again

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p alien-cli commands::init::tests --lib`

Supersedes the remaining unmerged part of #271. The chatbot example itself already landed in #428.

### DIFF
--- a/crates/alien-cli/src/commands/init.rs
+++ b/crates/alien-cli/src/commands/init.rs
@@ -32,6 +32,14 @@ struct TemplateInfo {
 
 const KNOWN_TEMPLATES: &[(&str, &str)] = &[
     (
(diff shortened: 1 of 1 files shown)


alongubkin:
Replaced by #483 after renaming the branch to satisfy the repository’s Linear-issue branch convention.

PR #493, opened 2026-08-25
feat: add CLI self-upgrade command

alongubkin:
## Summary

- add `alien upgrade` with a visible `alien update` alias
- stream and checksum-verify immutable binaries from the stable release channel before atomic replacement
- delegate npm and Homebrew installations to their package managers
- add `--dry-run` and `--force` safeguards

## Research

Reviewed Alien's release workflow, stable S3/CloudFront channel, installer routes, npm wrapper, Homebrew publishing, and public installation docs. Compared Codex, Claude Code, Deno, uv, and mise update behavior. The resulting model follows the common split: standalone installs self-replace; package-managed installs delegate to their owner.

## Validation

- `cargo check -p alien-cli --locked`
- `cargo test -p alien-cli commands::upgrade --lib`
- `cargo test -p alien-cli update_alias_parses_as_upgrade --lib`
- `target/debug/alien upgrade --dry-run`
- `target/debug/alien update --dry-run`
- npm delegation dry-run via `ALIEN_INSTALL_METHOD=npm`
- end-to-end local release server: streamed a checksum-tagged mock v99.0.0 executable, validated it, replaced a copied Alien binary, then executed the replaced path and confirmed `alien 99.0.0`

Linear: [ALIEN-577](https://linear.app/alienplatform/issue/ALIEN-577/add-a-self-update-command-to-the-alien-cli)

Companion docs PR follows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -374,6 +374,8 @@ dependencies = [
  "rand_distr",
  "reqwest 0.12.28",
  "rustls 0.23.43",
+ "self-replace",
+ "semver",
  "serde",
  "serde_json",
  "serde_yaml",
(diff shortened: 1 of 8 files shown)


alongubkin:
Production-channel verification passed after CI: copied the locally built PR binary to a temporary path, then ran the temporary binary with `alien upgrade --force` against the default `https://releases.alien.dev`. It fetched stable `v3.3.18`, verified/validated the production artifact, replaced only the temporary binary, and the replaced path executed successfully as `alien 3.3.18`. Binary comparison confirmed replacement: local debug build 554,816,488 bytes / SHA-256 `702592ed...`, installed production artifact 191,867,152 bytes / SHA-256 `49579f21...`. The user-installed CLI was not modified.

PR #498, opened 2026-08-26
fix: rejoin machines with rejected credentials

alongubkin:
## Summary

Normal `join` now repairs an installed external machine when its credentials were explicitly rejected by the control plane.

- Reads the installed machine service local status before returning the existing no-op result.
- Re-registers only for `authenticationFailed`; connected, degraded, and older services without local status keep the existing behavior.
- Keeps the machine ID and accepted machine plan, stops the service, removes only the revoked machine token, and restarts with the fresh join token.
- Waits for a new machine token before reporting success.
- Does not add a recovery flag or change normal leave behavior.

Keeping the machine ID is intentional: it lets the machine service use its normal cached-plan startup and preserve managed workloads while registration is restored.

## Validation

- `cargo test -p alien-deploy-cli commands::join::tests --lib` (42 passed)
- `cargo fmt -p alien-deploy-cli -- --check`
- `git diff --check`

The workspace Clippy command is currently blocked by pre-existing lint failures outside this change, including `result_large_err` throughout `alien-deploy-cli`.

Closes ALIEN-552.

### DIFF
--- a/crates/alien-deploy-cli/src/commands/join.rs
+++ b/crates/alien-deploy-cli/src/commands/join.rs
@@ -117,7 +117,7 @@ enum JoinAction {
     Reconfigure,
     Repair,
     Reregister,
-    ReregisterAndReinstall,
+    ReregisterThenRetry,
     NoOp,
 }
 
@@ -748,7 +748,7 @@ fn rejected_credentials_action(
         if install_state_matches_bundle_context(state, request, manifest) {
             JoinAction::Reregister
         } else {
-            JoinAction::ReregisterAndReinstall
+            JoinAction::ReregisterThenRetry
         },
     )
 }
@@ -773,7 +773,7 @@ async fn install_join(request: JoinRequest) -> Result<()> {
 
     if matches!(
         action,
-        JoinAction::Reregister | JoinAction::ReregisterAndReinstall
+        JoinAction::Reregister | JoinAction::ReregisterThenRetry
     ) {
         let state = read_install_state(&install_state_path(&paths))?;
         reregister_existing_join(&request, &state).await?;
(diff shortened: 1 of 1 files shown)


alongubkin:
Fixed the failure path in ded6276fb: after the service is stopped, any token-removal or service-install failure now attempts to restart the previous installed service before returning the original error. If the restart also fails, both failures are reported. Added regression tests for both recovery outcomes; all 44 join tests pass.

alongubkin:
Fixed in b70ba9de7. Re-registration now reads and validates the current token before stopping the service. If token removal or reinstall fails, it atomically restores the previous token before restarting the installed service; restoration and restart are both attempted and combined failures are reported. Added a regression test that removes the token and proves the restart observes the restored value. All 45 join tests pass.

alongubkin:
Fixed both findings in 306824730 by simplifying the recovery path. Re-registration no longer reinstalls or replaces the service definition: it stops the existing service, removes the rejected token, starts the same installed service, and waits for the new token. Token verification is inside the recovery boundary, so removal, start, and timeout/read failures all stop the attempt, restore the previous token atomically, and restart the previous service. All 45 join tests pass.

alongubkin:
Fixed in d19533ccf by separating credential recovery from bundle repair. Reregister now uses only the installed state: it updates the installed join-token file, preserves the installed executable/config/service definition, rotates the rejected machine token, and verifies the replacement. A bundle-version or bundle-URL difference no longer causes stale-manifest configuration to be written. A simultaneous control-plane URL change is rejected as unsafe. All 45 join tests pass.

alongubkin:
Fixed in e0d06b736. Rejected credentials now classify as either reregister (matching bundle context) or reregister-and-reinstall (changed version, URL, or service label). The combined path first restores valid registration through the installed service, then continues through the normal verified bundle installation. A control-plane URL change remains fail-closed. Added a regression test for the changed-bundle decision.

alongubkin:
Addressed the 4/5 summary in 652b5f9ae. A changed-bundle recovery no longer enters installation in the same transaction because the old service definition is not available for a safe rollback. The first join restores registration and returns an explicit instruction to rerun the same command; the second run sees healthy credentials and uses the existing normal bundle-upgrade path. This keeps recovery atomic and does not silently ignore the requested bundle.

PR #542, opened 2026-08-30
fix(cli): wait for hosted setup handoff

alongubkin:
## Summary

- wait for hosted release authority after setup reconciliation hands execution back to Platform
- preserve immediate corrective compute updates when failed deployments already retain release authority
- fail clearly on terminal lifecycle states or a bounded handoff timeout

## Test

- `cargo test -p alien-deploy-cli commands::up::tests -- --nocapture`


alongubkin:
Closing pending a first-principles correction: live staging shows the post-setup handoff can remain in provisioning with neither desired nor current release, so polling would only turn the API error into a timeout. The fix must preserve/restore release authority across setup reconciliation.

PR #544, opened 2026-08-31
fix: preserve durable setup metadata in Platform SDK

alongubkin:
## Summary
- regenerate the checked-in Platform OpenAPI artifacts from the authoritative Platform schema
- preserve durable setup revision metadata through generated reconciliation requests
- add a regression for the generated SDK serialization boundary

## Validation
- `cargo test -p alien-platform-api reconcile_state_preserves_runtime_update_metadata`
- repository generator is idempotent over the checked-in artifacts


alongubkin:
Superseded by the identical commit on a repository-conformant branch.

PR #556, opened 2026-09-01
feat: simplify setup link API and TypeScript SDK

alongubkin:
## Summary

- make setup-link targeting explicit while preserving the existing setup-link request fields
- add idempotent setup-link creation and direct entry points for gateway providers
- simplify TypeScript SDK request inputs so operation bodies are passed directly and workspace can come from the API key
- capture bounded provider error responses for internal diagnostics while keeping gateway responses sanitized
- regenerate the Platform TypeScript SDK from the updated OpenAPI contract

## Validation

- AI Gateway unit tests pass (90 tests)
- TypeScript SDK was regenerated from the source OpenAPI schema
- real-provider failure path verified: the upstream 400 body is captured internally while the caller receives the sanitized gateway error

## Merge order

Merge this PR before the corresponding Platform PR, which consumes the new AI usage diagnostic fields.


alongubkin:
@greptile

PR #573, opened 2026-09-03
feat(cli): enable the remote sandbox capability from the CLI

ItamarZand88:
Add `alien projects capabilities enable remote-sandbox --base-image <image> --max-session-lifetime-seconds 3600`, including JSON output and validation before authentication or project lookup. The CLI rejects missing flags, invalid image references, and session lifetimes above the API's eight-hour limit.

Updated against current `main`. Regenerated the API mirrors from the current API source and verified the TypeScript SDK. The API continues to support either a base image or an image bundle; this command uses the base-image path. The OpenAPI mirrors and TypeScript models now match `main`, so this PR makes no breaking API schema change.

Validation:

- API export and `pnpm run generate:platform-api` completed; the TypeScript SDK built successfully. Regeneration produced no model/method changes; incidental version and whitespace churn was discarded.
- TypeScript serialization checks passed for both supported image sources and the required session lifetime.
- Added CLI coverage for request serialization, invalid images, lifetime limits, and validation before authentication. All 248 CLI library tests pass, including 11 focused project-command tests. Formatting and lint checks pass.

- GitHub CI passed on `4a5a7c423c6a29e38067f4cfafbc12d0bb206fc2`: Rust crates, TypeScript packages and CLIs, and All tests passed.
- Greptile review: 5/5 on the current head.


alongubkin:
@greptileai Please re-review the current head `4a5a7c423c6a29e38067f4cfafbc12d0bb206fc2`. The stale TypeScript finding assumes the old version of this PR still removes `imageBundleUri` and requires `baseImage` in OpenAPI. That schema change has been removed.

The three API mirrors at this head are byte-for-byte identical to `main` (`65c4dc4eaf5f2c342486e1c1bb22451f485ce9ae`). `ConfigureRemoteSandboxRequest` has both `baseImage` and `imageBundleUri` as optional properties and only `maxSessionLifetimeSeconds` in `required`. Both image-source modes remain supported. The TypeScript model matches that contract. The CLI intentionally supplies `baseImage` and omits `imageBundleUri`.

I reran the API export and TypeScript SDK generator; there were no model or method changes versus main. Runtime serialization checks passed for both source modes and reject a missing lifetime. The 11 focused CLI tests also pass, including exact request serialization and offline validation.

The current diff against main contains only `crates/alien-cli/src/commands/platform/projects.rs`. Please evaluate that current contract rather than carrying forward the obsolete breaking-schema assumption.

PR #585, opened 2026-09-07
feat(sandbox): run a command as a job on any backend that has one

ItamarZand88:
Stacked on #584 → #583 → #582 → #581 → #580.

## Why the trait grows

A job is a command that outlives the request that started it. Three of the five sandbox backends already had the primitive — GCP runs *every* command as a job internally, and AWS and Kubernetes have the agent's `/v1/jobs/*` routes — but none of it was on the trait. A caller who needed a long command had to downcast to a concrete provider, which is the thing the trait exists to prevent.

`Sandbox` gains `start_job`, `poll_job`, `cancel_job`. `SandboxCapabilities` gains `jobs` so a caller branches before calling rather than on a refusal.

| Backend | `jobs` | Why |
|---|---|---|
| AWS, Kubernetes | true | delegate to the agent protocol |
| GCP Agent Platform | true | already ran commands as jobs; `run_command` is now built on the exposed path |
| Azure | false | runs no in-guest agent |
| Local | false | the local manager serves exec, files, preview and delete only |

## The classification is the contract

A start that goes unanswered is `SANDBOX_OUTCOME_UNKNOWN`, not retryable: the sandbox may have taken the command, and repeating it would run it twice. `start_job` routes through the same `send(…)` path that already carries that rule rather than reimplementing it.

Poll and cancel are repeatable and change nothing, so an unreachable sandbox stays `SANDBOX_UNREACHABLE` and retryable. GCP needed an explicit classifier for this: its generic path would have answered a standalone poll with `SANDBOX_COMMAND_FAILED`, which does not meet what the trait doc promises.

## Two deviations from the plan, both deliberate

**`poll_job` takes `Option<u64>`, not `since_seq: u64`.** The agent filters `seq > since` and sequences start at zero, so a bare `0` silently drops the first frame. `None` means "from the beginning", which also matches the agent wire and GCP's existing poll state.

**Refusal operations are named `sandbox.jobStart` / `sandbox.jobPoll` / `sandbox.jobCancel`**, matching the file-local convention (`sandbox.snapshot`, `sandbox.create`) rather than the plan's `start_job`. These strings are visible in the error payload, so flagging the choice.

## Testing

The test worth reading is the capability-faithfulness one: for every provider constructible in a test, `capabilities().jobs` is checked against whether `start_job` actually reached the transport or refused before it — observed through the transport, not through an error code, because Kubernetes returns `OperationNotSupported` for an unclaimed session too and a code-only check could not tell the two apart.

- `cargo test -p alien-core -p alien-bindings --features aws,gcp,azure,kubernetes --lib` — 301 + 565 pass.
- `cargo test -p alien-bindings --all-features --lib` — 330 pass. (`--all-features` without `--lib` pulls in the cloud integration suite, which needs a `.env.test`; it fails identically on an unmodified tree.)
- `pnpm --filter @alienplatform/bindings test` — 127 pass across 17 files. `tests/remote.test.ts` compares the TypeScript surface against the live napi class, so it confirms the addon really exposes the three methods.
- Schemas regenerated with `pnpm --filter @alienplatform/core generate`.


ItamarZand88:
Superseded by #587. The branch name needed its Linear id (ALIEN-686) for the PR-conventions check, and GitHub cannot repoint an open PR's head branch.

PR #586, opened 2026-09-07
feat(manager): accept sandbox-gateway as a gateway log source

ItamarZand88:
Stacked on #585. Its own diff is one file.

A gateway sends its per-request diagnostics to the customer's own Manager, under a telemetry token that names which gateway it is. `GatewayLogSource` is that name, and it gains a third variant.

The source is server-controlled: it travels inside `TelemetryCapability::GatewayLogs`, which is `deny_unknown_fields`, so a source has to be named here before Platform can mint a token carrying it. That is the property the new test pins from both directions — a minted `sandbox-gateway` source deserializes and round-trips to its stable OTLP string, and a source the Manager does not know fails to deserialize rather than reaching the log store.

`cargo test -p alien-manager auth` — 40 pass. The enum is not exposed in the Manager's OpenAPI, so no regeneration was needed.


ItamarZand88:
Superseded by #588, on ALIEN-687 for the same reason as #585.

PR #587, opened 2026-09-07
feat(sandbox): run a command as a job on any backend that has one

ItamarZand88:
## Summary

A job is a command that outlives the request that started it. Three of the five sandbox backends already had the primitive but none of it was on the `Sandbox` trait, so a caller who needed a long command had to downcast to a concrete provider.

Stacked on #584 → #583 → #582 → #581 → #580. What happens when a caller runs a long command:

1. The caller reads `SandboxCapabilities.jobs` and branches before calling, rather than calling and handling a refusal.
2. **`start_job` routes through the same `send(…)` path that already classifies an unanswered start as `SANDBOX_OUTCOME_UNKNOWN` and not retryable** — the sandbox may have taken the command, and repeating it would run it twice.
3. `poll_job` and `cancel_job` are repeatable and change nothing, so an unreachable sandbox stays `SANDBOX_UNREACHABLE` and retryable.

This PR changes running a long command from a downcast to a concrete provider into three methods on the trait, with a capability flag saying which backends have them.

## What was broken

Nothing was returning a wrong answer. The gap was that GCP ran *every* command as a job internally, and AWS and Kubernetes had the agent's `/v1/jobs/*` routes, but none of it was reachable through the trait — so a caller needing a long command had to downcast to a concrete provider, which is the thing the trait exists to prevent.

## What I did

- `Sandbox` gains `start_job`, `poll_job`, `cancel_job`. `SandboxCapabilities` gains `jobs`.
- Declared the capability honestly per backend:

| Backend | `jobs` | Why |
|---|---|---|
| AWS, Kubernetes | true | delegate to the agent protocol |
| GCP Agent Platform | true | already ran commands as jobs; `run_command` is now built on the exposed path |
| Azure | false | runs no in-guest agent |
| Local | false | the local manager serves exec, files, preview and delete only |

- Gave GCP an explicit classifier: its generic path would have answered a standalone poll with `SANDBOX_COMMAND_FAILED`, which does not meet what the trait doc promises.
- Two deviations from the plan, both deliberate. **`poll_job` takes `Option<u64>`, not `since_seq: u64`** — the agent filters `seq > since` and sequences start at zero, so a bare `0` silently drops the first frame; `None` means "from the beginning", which also matches the agent wire and GCP's existing poll state. **Refusal operations are named `sandbox.jobStart` / `sandbox.jobPoll` / `sandbox.jobCancel`**, matching the file-local convention (`sandbox.snapshot`, `sandbox.create`) rather than the plan's `start_job`; these strings are visible in the error payload, so flagging the choice.

## Files touched

- `crates/alien-bindings/src/traits.rs` — the three trait methods.
- `crates/alien-bindings/src/providers/sandbox/agent_protocol.rs` — the shared agent implementation behind AWS and Kubernetes.
- `crates/alien-bindings/src/providers/sandbox/gcp_agent_platform.rs` — `run_command` rebuilt on the exposed job path, plus the explicit classifier.
- `crates/alien-bindings/src/providers/sandbox/{aws,azure,kubernetes,local}.rs` — capability rows and refusals.
- `crates/alien-core/src/resources/sandbox.rs` — `SandboxCapabilities.jobs`.
- `crates/alien-bindings-node/src/sandbox.rs`, `packages/bindings/src/{types,factories,loader,index}.ts`, `packages/sdk/src/index.ts` — the TypeScript surface.
- Plus tests and 4 regenerated schema/Zod files — `pnpm --filter @alienplatform/core generate`.

## How I tested

- **Manually:** `pnpm --filter @alienplatform/bindings test` — 127 pass across 17 files. `tests/remote.test.ts` compares the TypeScript surface against the live napi class, so it confirms the addon really exposes the three methods rather than just declaring them in types.
- **Unit tests:** the one worth reading is the capability-faithfulness test in `crates/alien-bindings/src/providers/sandbox/jobs_capability_tests.rs`: for every provider constructible in a test, `capabilities().jobs` is checked against whether `start_job` actually reached the transport or refused before it — observed through the transport, not through an error code, because Kubernetes returns `OperationNotSupported` for an unclaimed session too and a code-only check could not tell the two apart. `cargo test -p alien-core -p alien-bindings --features aws,gcp,azure,kubernetes --lib` — 301 + 565 pass. `cargo test -p alien-bindings --all-features --lib` — 330 pass.
- **Anything I couldn't test:** `--all-features` without `--lib` pulls in the cloud integration suite, which needs a `.env.test`; it fails identically on an unmodified tree. No job was run against a live sandbox on any cloud.

This is a trait-surface change rather than a permissions, token or route change, so there is no attack list. The one safety property worth naming is double execution: an unanswered `start_job` classifies as `SANDBOX_OUTCOME_UNKNOWN` and is not retryable, because the sandbox may have taken the command — `start_job` routes through the existing `send(…)` path that already carries that rule rather than reimplementing it.


ItamarZand88:
@greptile review

PR #594, opened 2026-09-08
fix: require explicit subnets for postgres setup

alongubkin:
## Problem
Setup offered the account default network for a Frozen Postgres resource, although its private networking needs explicit subnet IDs. Generated expressions could resolve to AWS::NoValue and fail during stack creation.

## Changes
Include setup-owned Postgres in the shared named-subnet requirement. Keep runtime-owned databases and Kubernetes targets unaffected. Describe the restriction in terms of private resources rather than only sandbox connectors.

## Local validation
- New core regression covers Frozen, Live, Kubernetes and empty stacks.
- 48 relevant sandbox generator tests passed across local runs, including cfn-lint and Terraform plan rejection for use-default.
- Two additional existing sandbox tests fail because installed cfn-lint does not recognize MicroVM IAM actions; no suppressions added.
- Updated generated snapshots were reviewed (description text only).
- No cloud resources or installations changed.

ALIEN-735

### DIFF
--- a/crates/alien-cloudformation/src/generator.rs
+++ b/crates/alien-cloudformation/src/generator.rs
@@ -1089,9 +1089,9 @@ fn add_network_parameters(
     // parameter is checked before the transform runs, where a template Rule is not yet reached.
     let restricted = restricts_network_mode(stack, target);
     let description = if restricted {
-        "Choose create-new for a managed VPC, or use-existing for your VPC. A sandbox in this \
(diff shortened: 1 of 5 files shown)


alongubkin:
Companion emitter fix: https://github.com/alienplatform/platform/pull/617. Tested together with 26 passing Postgres generator tests, including CloudFormation lint and Terraform validate for compute + Postgres. Coordinate deployment with the companion to avoid an undeclared optional-network condition in compute templates.

PR #607, opened 2026-09-09
fix: continue health observations after refresh failure

alongubkin:
## Summary

Continue observation after a deployment health check fails, without resetting resource controllers or automatically retrying provisioning/update/deletion.

- Heartbeat acquisition includes `refresh-failed`; SQLite only acquires these without retry for a heartbeat filter that also requests `running`.
- The heartbeat runner executes one observation from either status instead of stopping before the failed-state handler.
- A refresh failure with no retry request uses the existing observation path. It returns to Running only after resource health recovers. Explicit resource retry remains a separate path.

## Evidence and validation

A disposable, previously healthy single-host service was subjected to host loss. Automatic replacement restored the endpoint; a fresh Git clone matched its original commit and stored PR/comment data survived. The deployment nevertheless remained refresh-failed for more than eight minutes after failure detection. Source inspection found the running-only heartbeat filter, pre-step failed-status gate, and retry-only refresh-failure handler.

Added regressions for SQLite acquisition isolation, the runner's failed-status entry, and an actual test controller's healthy → failed observation → recovered observation sequence. The latter checks that the existing controller survives (second observation rather than a recreated controller) and no retry or release change occurs.

Formatting and diff checks pass. Rust tests are pending CI; a new local manager build is not feasible with the available disk space. This PR does not claim live acceptance yet.

Release acceptance: repeat host loss, verify failure visibility, autonomous status recovery without retry, unchanged applied release and retained application data. Also verify explicit retry behavior for provisioning/update/delete failures remains unchanged.

### DIFF
--- a/crates/alien-deployment/src/lib.rs
+++ b/crates/alien-deployment/src/lib.rs
@@ -296,14 +296,7 @@ pub async fn step(
             }
         }
         DeploymentStatus::RefreshFailed => {
-            running::handle_refresh_failed(
-                current,
-                require_target_stack()?,
-                config,
-                client_config,
-                service_provider,
-            )
-            .await?
+            running::handle_refresh_failed(current, config, client_config, service_provider).await?
         }
         DeploymentStatus::Deleted => {
             debug!("Deployment is deleted, no action");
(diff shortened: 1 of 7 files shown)


alongubkin:
@greptileai please review the latest head, including observation without a pending desired release. CI is running; live acceptance will verify automatic recovery of the existing stale health failure, without a retry.

PR #612, opened 2026-09-10
fix(manager): thread operations-plugin sync reports through the sync loop

ab-alien-dev:
## Summary

- The Operator's `SyncRequest` already carries `operations_report`, and `alien-core::sync::SyncResponse` already has `target_operations_bundle_set` — but `alien-manager`'s own HTTP-layer types (`ReconcileRequest`/`AgentSyncRequest` for input, `ReconcileResponse`/`AgentSyncResponse` for output) never had matching fields. The report was silently dropped before it ever reached a `DeploymentStore` implementation.
- Added `operations_report` to `ReconcileData` and `AgentSyncRequest`.
- Added `target_operations_bundle_set` to a new `ReconcileOutcome` (replacing `DeploymentStore::reconcile`'s `DeploymentRecord` return type) and to `AgentSyncResponse`, threading both through every call site (`routes/sync.rs`, `loops/deployment.rs`, `transports/manager.rs`, the SQLite store).
- Push-mode paths pass `operations_report: None` — that's a pull-mode-only concept.

## Test plan

- [x] `cargo build -p alien-manager -p alien-operator -p alien-core` — builds clean standalone
- [x] `cargo build --workspace` — no errors
- [x] Verified end-to-end against a live local Kubernetes Operator: enabling/disabling operations plugins now correctly reports the loaded bundle hash back through `DeploymentStore::reconcile`


ab-alien-dev:
This review is stale — last reviewed commit `1ae8aea3` (the first commit on this PR). Both findings were fixed in `7f5bc47e2`, pushed shortly after: `OperationsReport`/`TargetOperationsBundleSet` and the `AgentSyncRequest.operationsReport`/`AgentSyncResponse.targetOperationsBundleSet` fields are now registered in `alien-manager`'s `api.rs` and present in the regenerated `crates/alien-manager/openapi.json`, and the four new types are registered in `alien-core`'s `schema_exporter.rs`.

lilienblum (changes requested):
Requesting changes for a cross-PR correctness blocker: OperationsSyncHandlerHook takes no resolved operator configuration, so the platform implementation reconstructs the data directory from DATA_DIR or /var/lib/operator and ignores the CLI --data-dir value. The real-operator demo launches alien-operatorx with --data-dir and no DATA_DIR, which separates the operator database/target state from the synced bundle directory and may try to write to an unavailable /var/lib/operator. Please pass the resolved OperatorConfig or data-directory path into the hook and construct the sync handler from it.

crates/alien-operator/src/cli.rs
```diff
@@ -168,10 +168,17 @@ pub type AccessRequestSyncLoopHook = fn() -> Option<std::sync::Arc<dyn AccessReq
 /// so the OSS operator runs no authorized operations commands.
 pub type OperationsExecLoopHook = fn() -> Option<std::sync::Arc<dyn OperationsExecLoop>>;
 
+/// Optional fifth hook that lets downstream binaries inject a real
+/// [`OperationsSyncHandler`] (downloads/hot-reloads the plugin registry
+/// toward the sync loop's target bundle set and reports the loaded catalog
+/// back). Defaults to `None`, so the OSS operator reports nothing.
+pub type OperationsSyncHandlerHook = fn() -> Option<std::sync::Arc<dyn OperationsSyncHandler>>;
```
lilienblum, line 175: [blocking] A zero-argument hook cannot use the resolved --data-dir. The downstream implementation is forced to reread environment defaults and ignores CLI-only configuration, causing sync state and bundles to use a different directory. Pass the resolved OperatorConfig or data_dir into this hook, then invoke it after parsing with that value.

ab-alien-dev: Fixed in 08394c770: `OperationsSyncHandlerHook` now takes `data_dir: &str`, and the CLI passes `operator_config.data_dir` (the CLI's fully-resolved value — `--data-dir`, then `DATA_DIR`, then the built-in default, in that order) into the hook at the call site in cli.rs, instead of the hook re-deriving it independently.

lilienblum (approved):
Re-reviewed at d29b30cd3b14cb4bca81f8751aa6e4e092e2f51d. No material findings. The previous blocker is fixed: OperationsSyncHandlerHook now receives the CLI-resolved OperatorConfig data directory, and the paired platform revision uses it. Rust CI is green; locally, alien-core sync tests passed (11/11) and alien-operator tests passed (55/55).

PR #614, opened 2026-09-11
fix: allow deployment credentials to delete their deployment

alongubkin:
## Summary
- allow a retained deployment-manager credential to delete only its own deployment
- keep deployment-group bootstrap credentials unable to delete deployments
- preserve workspace and deployment scope checks

## Why
Terraform and other long-lived setup tools retain deployment-scoped authority after their broader bootstrap token expires. That credential must own the complete lifecycle of its exact deployment, including teardown.

## Validation
- focused OSS authorization regression passed
- real local Terraform teardown reached the hosted manager with the retained deployment credential; cross-deployment and deployment-group denial remain covered

Stacked on #613.


alongubkin:
@greptileai review

PR #620, opened 2026-09-14
fix: retry failed generated CLI teardown

alongubkin:
## Summary

- request a new delete attempt when setup-owned teardown previously failed
- keep `teardown-required` idempotent while work is still active
- preserve the tracked deployment until teardown succeeds

## Validation

- `cargo test -p alien-deploy-cli failed_setup_teardown_requests_a_new_delete_attempt`
- real generated CLI retry against a failed AWS setup: the API re-armed the same delete operation, the CLI resumed from the failed Aurora delete state, removed the remaining IAM resources, and completed successfully
- verified no IAM roles, IAM policies, Aurora clusters, Aurora instances, or DB subnet groups remained for the deployment prefix

Linear: ALIEN-793

PR #621, opened 2026-09-14
fix: renew setup teardown leases

alongubkin:
## Summary

- reuse the shared lease-renewal guard around setup-owned teardown
- cancel setup teardown before ownership can expire when renewals fail
- cover lease renewal during an operation longer than the renewal interval

## Root cause

Setup-owned teardown has its own long-running resource loop. Unlike the normal deployment runner, it did not renew the deployment lease. A real generated-CLI AWS destroy therefore lost its claim after exactly five minutes while Aurora was still deleting.

## Validation

- `cargo test -p alien-deployment --lib` (103 passed)
- Real pre-fix reproduction: deployment `dep_timzeyh4w5ia0owdlytoow3wes0k` acquired at 06:45:20 UTC, lease expired at 06:50:20.625, and its next checkpoint returned `DEPLOYMENT_UPDATE_CLAIM_LOST` while RDS remained in `deleting`.
- Supported retry then completed teardown; exact-prefix AWS and Alien metadata audits were empty. A post-fix >5-minute real teardown remains the final acceptance gate.

## Stack

Stacked on #620 because the generated CLI retry behavior is needed to recover the pre-fix failure fixture cleanly.

Linear: ALIEN-794


alongubkin:
@greptileai Please review the current head, especially cancellation safety and reuse of the existing lease-renewal contract.

alongubkin:
@greptileai

alongubkin:
@greptileai

PR #622, opened 2026-09-14
fix(cli): send uploadId on operations plugin publish

ab-alien-dev:
## Summary
- The embedder's presigned-upload endpoint for custom operations plugin bundles now mints a one-time upload id per call and keys the stored object by it, rather than by plugin name/version, so no upload can ever overwrite bytes an already-published version's catalog row still points at.
- Updates `alien operations publish` to capture that upload id from the upload-url response and pass it back on the publish call, matching the updated contract.
- Resyncs the vendored OpenAPI spec copies (`client-sdks/platform/openapi.json` and the two Rust variants) to the current schema.

## Test plan
- [x] `cargo build -p alien-cli` succeeds
- [ ] CI green


ab-alien-dev:
The TypeScript Platform SDK doesn't need regeneration for this change — the affected flow (`alien operations publish`) is Rust-only, calling the platform API directly via `reqwest` in `crates/alien-cli/src/commands/operations/platform_actions.rs`. It doesn't go through `client-sdks/platform/typescript` at all. The checked-in TypeScript SDK is a separate consumer surface for other integrations; it's out of scope for this PR and can be regenerated independently whenever a TypeScript caller needs the new `uploadId` field.

ab-alien-dev:
@greptileai review

PR #625, opened 2026-09-14
test: use resource-scoped operations permission

alongubkin:
## Summary
- make the operations permissions test use a real permission set with the resource binding the command requires
- keep stack-only permission sets correctly rejected by the production generator

## Root cause
The test selected the first permission set with any AWS configuration. `observe/observe` is valid but stack-only, while `alien operations permissions` intentionally generates a policy for `BindingTarget::Resource`. Registry ordering therefore made the test exercise an unsupported scope.

## Validation
- `cargo test -p alien-cli commands::operations::permissions::tests --lib` (4 passed)
- `git diff --check`

Linear: ALIEN-798


alongubkin:
@greptileai please review the current head.

PR #635, opened 2026-09-17
revert: restore previous OTLP exporter behavior

alongubkin:
Plain revert of #632. The exporter upgrade lost HTTPS support through a new HTTP-client dependency, and the 750ms timeout regressed otherwise healthy slower receivers. Restore the prior dependency versions, timeout, and shutdown behavior; any retry changes will be proposed separately.

Validation: the complete resulting Git tree exactly matches the parent of e93005722acd7a4a0b09f4e489e56e472a5638e8. No replacement implementation or new tests are mixed into this rollback. Full CI is not being waited on for the requested admin squash merge.

ALIEN-850


alongubkin:
Admin squash merge requested by the repository owner to roll back this unreleased change quickly. Verified that this PR restores the exact complete tree immediately preceding e93005722acd7a4a0b09f4e489e56e472a5638e8; no new behavior is included. Full CI has not completed and is not claimed green. Follow-up fixes will be reviewed separately; this merge is not a deployment.

PR #636, opened 2026-09-17
fix: preserve first-party deployment context

alongubkin:
## Background

`alien deploy --channel <name>` did not carry one consistent deployment context through the first-party setup flow:

- the requested release channel was omitted from deployment creation and setup-input validation;
- manager discovery could use an install/user credential, and deployments without an explicit workspace accidentally reused that credential for provisioning instead of switching to the deployment-scoped token.

The result was channel drift (a staging deploy could validate or create against production) and authorization behavior that changed depending on whether `--workspace` was present.

## Reproduce

1. Configure different inputs on the production and staging release channels.
2. Run `alien deploy --channel staging` and provide a staging-only input.
3. Before this change, inspect the create and first-party-input request bodies: neither includes `staging`.
4. Run the same deploy without an explicit workspace. Before this change, provisioning reuses the discovery client rather than authenticating with the new deployment token.

## Fix

- Carry `releaseChannel` in both deployment creation and first-party input requests.
- Always construct the provisioning manager client with the deployment-scoped token; attach the workspace header only when available.
- Update the checked-in API schemas and generated TypeScript request model.
- Add a focused regression proving `staging` survives both request-building paths.

The server field is backward compatible: clients that omit it continue to use production.

## Validation

- `CARGO_BUILD_JOBS=2 CARGO_INCREMENTAL=0 CARGO_PROFILE_DEV_DEBUG=0 cargo nextest run -p alien-cli first_party_creation_preserves_requested_release_channel`
- Result: 1 passed, 282 skipped
- `git diff --check`

Linear: ALIEN-852


alongubkin:

alongubkin:
@greptileai Please re-review the latest head; the generated client now compiles with the synchronized 404 contract.

alongubkin:

alongubkin:
@greptileai please re-review the latest head; CI exposed and this commit fixes the stale OpenAPI error-schema reference introduced during conflict resolution.

alongubkin:
@enclave-ai please re-run the security analysis on the current head; the existing check has remained in progress for over five hours.

PR #637, opened 2026-09-17
fix: plan configurable cloud storage

alongubkin:
## Background

The instance catalog stores two different kinds of storage in one numeric field:

- fixed local/NVMe capacity on storage-optimized machines;
- a conservative default size for provider disks on ordinary cloud machines.

The compute planner treated both as a hard hardware maximum. A workload requesting 100 GiB was therefore rejected on an otherwise suitable `m8i.2xlarge` because its catalog default is 20 GiB—even though the cloud controller provisions that disk and can size it to 100 GiB. Requests above the largest fixed-local catalog entry also failed while choosing a recommendation, before an explicitly selected configurable machine could be considered.

## Reproduce

1. Create a container pool requiring 100 GiB ephemeral storage.
2. Select AWS `m8i.2xlarge` (including the nested-virtualization case).
3. Before this change, selection validation reports that the machine does not satisfy the pool because `20 GiB < 100 GiB`.
4. Increase the request beyond the largest storage-optimized catalog entry (for example 8,000 GiB). Before this change, recommendation selection fails outright even though a provider disk can be configured to that size.

## Fix

- Explicitly distinguish configurable provider storage from fixed local storage in the catalog API.
- Apply the catalog capacity ceiling only to storage-optimized/fixed-local machines.
- Prefer fixed-local NVMe for large requests, then fall back to an ordinary machine with configurable storage when no fixed-local entry is large enough.
- Materialize configurable disk profiles as `max(provider default, workload request)` so the controller receives the requested capacity rather than silently falling back to 20 GiB.
- Use the same effective profile in recommendations, machine options, and the preflight mutation.

Fixed-local capacity remains a hard ceiling; this PR does not silently reinterpret local NVMe capacity.

## Validation

- planner regression: 100 GiB remains valid on `m8i.2xlarge`, and the machine option reports 100 GiB
- catalog and end-to-end planner regressions: an 8,000 GiB request falls back from fixed-local NVMe to the explicitly selected configurable machine and preserves 8,000 GiB
- preflight regression: the materialized capacity-group profile carries 100 GiB
- focused local `cargo nextest` runs pass; rustfmt and `git diff --check` pass

Linear: ALIEN-855


alongubkin:

alongubkin:
@greptileai Please re-review the latest head. The configurable-storage fallback now derives the vCPU policy from the effective candidate family, with a high-demand regression proving the 8-vCPU cap and horizontal scaling.

alongubkin:

PR #638, opened 2026-09-17
fix: validate cloud config before deployment creation

alongubkin:
## Background

`alien deploy` created or registered the deployment first, persisted its deployment-scoped token in the runner-local tracker, and only then loaded the target cloud configuration. A basic local prerequisite error such as a missing AWS region therefore left a durable pending deployment even though no cloud operation could begin. This is especially painful in ephemeral CI because the runner that holds the tracker may disappear.

## Reproduce

1. Use a fresh runner with valid Platform/project authentication but no usable AWS region configuration.
2. Run `alien deploy --platform aws ...` for a new deployment.
3. Before this change, the Platform deployment record and token are created first; `ClientConfig::from_std_env(Platform::Aws)` then fails while provisioning begins.
4. A retry from another runner must discover/recover or delete the orphaned pending record.

## Fix

Build and validate the provider client immediately after resolving the requested platform, before tracker lookup, deployment-group session creation, deployment creation, or token persistence. The validated client is retained and reused by the provisioning loop.

Machines remains exempt because it creates a join command and does not use a runner-local cloud client. Local development still exits into its dedicated path before this validation.

This is local feasibility validation only; authoritative account/project/subscription checks and cloud ownership protections remain unchanged.

## Validation

- `cargo check -p alien-cli`
- focused deploy tests: 7 passed, 275 skipped
- rustfmt and `git diff --check`

Linear: ALIEN-853


alongubkin:

alongubkin:

PR #642, opened 2026-09-17
fix: preserve semantic deployment failures through finalization

alongubkin:
## Background

The deployment step loop represents controller and preflight failures as a successful Rust return containing `LoopOutcome::Failure`. The detailed failure is durably checkpointed in `DeploymentState.error`.

Every CLI caller then performs a final reconcile/release. Because the operation was still represented as `Ok(RunnerResult)`, a failure in that finalization replaced the real controller error.

## Reproduction

1. Start a deployment whose compute pool has no selected machine.
2. The runner checkpoints the useful error (for example, `compute pool 'preview' has no selected machine`) and stops with `LoopOutcome::Failure`.
3. Make the final reconcile fail (the observed case was an execution attempt that was no longer live).
4. The CLI reports only the finalization error, hiding the actionable root cause.

## Root cause

The runner has two failure channels: transport/runtime errors use `Err`, while terminal semantic failures use `Ok(RunnerResult { outcome: Failure })` plus `DeploymentState.error`. The finalization combiner only understood the first channel.

## Fix

- Convert terminal semantic failures back to operation errors before finalization in all deploy and destroy entry points.
- Keep the operation error as the top-level error when operation and finalization both fail.
- Preserve the finalization error under structured `context.finalizationError`.
- Keep successful and transport-error behavior unchanged.

## Verification

- `cargo test -p alien-deployment semantic_runner_failure_returns_the_checkpointed_operation_error --lib`
- `cargo test -p alien-deployment operation_error_is_retained_when_finalization_also_fails --lib`

Both pass locally.


alongubkin:

alongubkin:
Superseded by #643: identical fix on the required ALIEN-628 branch naming convention.

PR #644, opened 2026-09-17
perf: generate only used Platform API operations

alongubkin:
## Summary

- generate the Platform Rust client from the 58 operations used by Alien instead of all 229 operations
- retain transitively reachable components, discriminator mappings, path metadata, and security schemes
- preserve the complete published SDK behind the default `full-api` feature
- normalize uniquely matching nullable enum copies to named components so Progenitor does not emit duplicate Rust types
- fail generation on missing operations, duplicate operation IDs, unresolved component references, or more than 600,000 filtered generated lines

## Impact

- generated Rust: 1,398,184 -> 534,989 lines (-61.7%)
- generated Rust size: about 56 MB -> 22.2 MB (about -60.4%)
- generated structs: 5,533 -> 2,003 (-63.8%)
- generated enums: 3,663 -> 1,372 (-62.5%)
- generated impls: 38,566 -> 14,187 (-63.2%)
- Platform check unit: 326.5s baseline -> 35.4s observed
- Platform build script: 30.1s baseline -> 7.2s observed
- empty-target `cargo check -p alien-cli --timings`: 7m13s baseline -> 2m34s observed

The timing runs used empty Cargo target directories on the same machine but were recorded under different system load and filesystem-cache conditions. The generated-source reduction is deterministic; timing ratios are observed rather than a perfectly controlled attribution.

## Validation

- `cargo check -p alien-cli`
- `cargo test -p alien-platform-api --no-default-features --features rustls`
- `cargo test -p alien-platform-api`
- generated full SDK still exposes all 229 operations
- synthetic coverage for transitive references, cycles, discriminator mappings, path metadata, security schemes, missing/duplicate operations, unresolved references, and nullable enum normalization

Linear: ALIEN-861


alongubkin:
@greptileai The P1 packaging finding is fixed: the crate include list now contains `build/**/*`, and `cargo package -p alien-platform-api --list --allow-dirty` confirms `build/openapi_filter.rs` is present. Please re-review.

PR #647, opened 2026-09-17
perf: isolate permission types from core

alongubkin:
## Summary

- move the unchanged permission data model into a small serde/indexmap-only crate
- re-export every type through `alien_core::permissions` to preserve public paths and type identity
- validate permission JSONC in the build script without compiling all of `alien-core` for the host

## Impact

Controlled empty-target `cargo check -p alien-permissions --timings`:

- wall time: 16.09s -> 11.69s (-27.3%)
- host `alien-core`: 8.2s -> removed
- replacement host types crate: 0.2s
- cold target size: 751 MB -> 392 MB

The target `alien-core` check remains and compiled once, as expected.

## Compatibility

- original and extracted model source is identical apart from module documentation
- `alien_core::permissions::*` and crate-root re-exports remain available
- `alien-core/openapi` forwards schema derivation to the extracted types
- typed JSONC parsing plus empty-ID, empty-description, and duplicate-ID build failures remain intact

## Validation

- `cargo test -p alien-permissions` (159 tests plus docs)
- `cargo test -p alien-core --lib permissions`
- `cargo check -p alien-core --features openapi`
- `cargo test -p alien-permission-types --features openapi`
- `cargo package -p alien-permission-types --allow-dirty --no-verify --list`
- `cargo tree -p alien-permissions --edges build` contains no `alien-core`
- `git diff --check`

Linear: ALIEN-863


alongubkin:
@greptileai The P1 release finding is fixed: `alien-permission-types` is now in the explicit crate publication list immediately before `alien-core`. Both package file enumerations pass, and an order check confirms the dependency precedes its consumer. Please re-review.

PR #651, opened 2026-09-17
feat(sandbox): publish the GCP sandbox agent image

ItamarZand88:
## Summary

Nothing here built a container image for the sandbox agent, so a GCP Agent
Platform sandbox had no image to pull. This adds one, publishes it beside the
five images the release workflow already ships, and makes the image contract a
single definition that both sandbox Dockerfiles are rendered from.

When a GCP sandbox starts:

1. Agent Platform pulls `ghcr.io/alienplatform/alien-sandbox-agent` and runs it
   as uid 1000 on port 8080, both declared by the environment template.
2. **The agent reads the rest of its contract from the image's own `ENV` and
   refuses to start without any of it.** The template deliberately carries no
   environment, because the supervised command shares the agent's uid and could
   read whatever were placed there.
3. `transport` serves an uncapabilitied request only from a socket it cannot
   trace back to that uid, which keeps the sandboxed code out.

This takes the sandbox image contract from values repeated across five files to
one definition the Dockerfiles are generated from.

## What I did

Added `docker/Dockerfile.alien-sandbox-agent`, modelled on the operator image:
multi-arch selector stage, `git` for the sandboxed command, the agent root-owned
and unwritable by the uid it supervises, and a session directory only that uid
can write.

Wired it into the release workflow beside the existing images, and added a smoke
test that qualifies the published image on both architectures: the ELF
`e_machine` it actually contains, the exec identity, that the supervised uid
cannot unlink its own supervisor, that an invalid config is refused by name, and
a real run that reaches its listener and is still up afterwards.

Moved the image contract into `alien-core::sandbox_image` as data. The GCP
Dockerfile is now generated output with one test that regenerates and compares
the whole file, which replaced a 261-line lexical scanner that could only pin
what someone thought to assert. Isolation and the trailing `USER` line are one
enum, so an image cannot declare a mode its ending contradicts.

## How I tested

Built and ran the image on both platforms and read the result back: the six
`ALIEN_SANDBOX_*` values, `USER 1000:1000`, the exposed port, the entrypoint,
`/sandbox` at `drwx------`, and the agent `root root -rwxr-xr-x`.

Proved the AWS Dockerfile is byte-identical across the refactor by building the
renderer at both commits and diffing 84 rendered outputs, including a base image
containing literal braces. `shasum -a 256` matches on both sides.

Ran the artifacts a customer installs under `INSTA_UPDATE=no`: CloudFormation
16/16 and the Terraform sandbox snapshot with a real `terraform validate`. No
snapshot moved. Full suites green across all eight crates the change touches.

Broke the drift test four ways that the old scanner passed silently, and
confirmed each now fails: `/sandbox` at 0777, the login shell changed to
`/bin/sh`, the passwd append redirected to `/dev/null`, and a later `chmod 0777`
on the agent binary.

Walked the security questions this change raises:

- The supervised command shares the agent's uid under `platform` isolation, so
  it could answer in the agent's place. It cannot: `peer::transport_may_serve`
  refuses any in-guest socket whose owner is the exec uid, and refuses the mode
  outright where `/proc/net/tcp` is unreadable.
- It could rewrite its own supervisor. It cannot: the agent is `0:0 0755` in a
  root-owned directory, and the smoke test proves the `rm` is denied on a real
  container rather than trusting the mode bit.
- A capability could leak through the template. Nothing is placed there to leak;
  the template carries no environment at all, by design.
- The image could ship the wrong architecture. The ELF header is read off the
  binary, because once binfmt is registered running it proves nothing.

Nothing turned up.


ItamarZand88:
Valid, and fixed in c70ba88. The harness covers every failure mode of the release smoke test, and nothing ran it, so a change to `smoke-sandbox-agent.sh` alone reached no CI at all.

Added `scripts/smoke-sandbox-agent*` to the rust paths filter and a step that runs `scripts/smoke-sandbox-agent.test.sh`. It drives the script against a stub docker once per failure mode and asserts each one exits nonzero **and** prints an `::error::` annotation, so a mode that dies silently fails the job. It needs no daemon and no image, and runs in about 36 seconds.

Scoped the filter to the harness files rather than all of `scripts/**`, so an unrelated script edit does not pull in the whole Rust job.

ItamarZand88:
@greptile review

PR #653, opened 2026-09-17
perf: compile GCP Compute models in parallel

alongubkin:
## Summary

- move the large handwritten GCP Compute DTO set into a dependency crate
- re-export the DTOs from the existing `compute` module so public source paths remain compatible
- let Cargo compile the HTTP client and model-heavy crate in parallel

## Measured impact

The complete stacked CLI check improved from 99.00s after the Azure sharding change to 95.25s with this change: **3.75s / 3.8%**. This is the primary workflow result.

Focused, controlled cold `cargo check -p alien-gcp-clients --timings` attributes the improvement:

- before: 44.07s wall / 74.79s user
- after: 23.93s wall / 74.22s user
- focused wall-clock reduction: 45.7%

The nearly unchanged CPU time shows critical-path parallelism rather than removed validation or functionality. The focused percentage is not presented as a full-workflow improvement.

## Validation

- 42 GCP library tests pass
- every GCP integration-test binary compiles with `cargo test --no-run`
- all-platform `alien-infra` check passes
- package manifests and release ordering validate
- mechanical source comparison confirms DTO definitions are unchanged; only the private HTTP helper remains in the parent crate

The credential-gated Compute integration suite requires `.env.test`; all 17 tests fail immediately at that existing environment precondition when credentials are absent.

Linear: ALIEN-869


alongubkin:
Closing after the complete-workflow audit. The focused GCP client check improved 44.07s → 23.93s, but the controlled complete CLI stack improved only 3.75s / 3.8%. That does not justify another generated-crate boundary, publication ordering, and ongoing maintenance for this campaign. The experiment and measurements remain documented; we are prioritizing changes with material required-workflow impact.

PR #655, opened 2026-09-17
perf: compile Azure infrastructure controllers in parallel

alongubkin:
## Summary

- move the complete Azure controller/importer/permissions slice into `alien-infra-azure`
- add narrow Azure provider and permissions service boundaries through the typed registry
- preserve aggregate public paths and persisted controller type/version compatibility
- register sibling-crate trait instances explicitly in controller tests

This PR is stacked on #654.

## Evidence

- empty-target `cargo check -p alien-infra --all-features --timings`: **88.85s**, down from **91.84s** after the AWS split (**3.3%**)
- Azure checks in **4.2s** in parallel; the remaining aggregate falls from **13.12s** to **9.1s**
- warmed Azure worker edit: **8.18s**
- `cargo check -p alien-infra-azure`
- `cargo check -p alien-infra --all-features`
- `cargo check -p alien-infra-azure --features test-utils --tests`
- all **126** `alien-infra-azure` library tests pass
- all **107** AWS sibling tests pass on the stacked foundation after the typed test-service fix

## Compatibility

Existing aggregate module paths are compatibility re-exports. Persisted controller tags and `_controllerStateVersion = 1` are unchanged. Production service registration remains in `alien-infra`; the sibling test adapters are `cfg(test)` only.

Linear: ALIEN-876

PR #656, opened 2026-09-17
perf: compile GCP infrastructure controllers in parallel

alongubkin:
## Summary

- move the 20,161-line GCP controller/importer slice into `alien-infra-gcp`
- preserve the existing `alien_infra` public paths through compatibility re-exports
- adapt the platform provider through a narrow 14-client GCP service interface
- preserve persisted controller tags and controller state version 1

Stacked on #655. Tracks ALIEN-877.

## Evidence

- all 89 `alien-infra-gcp` library tests pass
- standalone GCP, aggregate all-feature, and all GCP test targets check successfully
- controlled sequential empty-target CLI benchmark: 128.95s exact parent vs 127.88s candidate (1.07s / 0.8% faster)
- remaining aggregate rustc unit: 9.1s after the Azure split vs 5.84s here
- warmed GCP worker edit through the aggregate: 5.90s

The cold gain is deliberately described as modest. The larger benefit is provider-local incremental iteration and completing the cloud-controller decomposition.

## Validation

- `cargo check -p alien-infra-gcp`
- `cargo check -p alien-infra --all-features`
- `cargo check -p alien-infra-gcp --features test-utils --tests`
- `cargo test -p alien-infra-gcp --features test-utils --lib`
- `cargo metadata --no-deps --format-version 1`
- `git diff --check`


alongubkin:
Closing because the controlled complete-CLI improvement is 1.07s / 0.8%, below the maintenance threshold for a 72-file provider boundary and another published crate. The experiment is valid and the incremental GCP edit result is useful evidence, but compilation time alone does not justify this architecture.

PR #658, opened 2026-09-17
fix: preserve compute selections in public deploy

alongubkin:
## Background

Public `alien deploy` could not express the compute choice that ComputeCluster preflight requires. The generated installer already accepts `ComputeSettings`, but the public config rejected `compute` as unknown and both public creation request paths omitted it.

Closes ALIEN-882.

## Reproduction

```toml
name = "production"
platform = "aws"

[compute.pools.preview]
mode = "fixed"
machines = 1
machine = "m8i.2xlarge"
```

Before this change, `alien deploy --config deployment.toml` fails parsing with `unknown field compute`.

## Root cause

`DeployConfigFile` and `ResolvedDeployArgs` omitted the core compute type. The first-party JSON request never serialized `stackSettings.compute`, while deployment-group SDK creation hard-coded `compute: None`.

## Fix

- Accept the canonical core `ComputeSettings` in public deploy TOML.
- Validate every pool selection locally before creating a deployment.
- Preserve compute in both first-party JSON and deployment-group SDK request paths.
- When resuming, accept an exact match but reject conflicting requested compute instead of silently ignoring it.

This does not change generated-installer behavior or infer billable capacity.

## Verification

- `cargo check -p alien-cli`
- `cargo test -p alien-cli compute_ --lib`
- Focused tests cover fixed selection serialization and invalid autoscale bounds before any API request.


alongubkin:
@greptileai review

alongubkin:
@greptileai re-review

PR #659, opened 2026-09-17
fix: surface safe JSON command failures on stderr

alongubkin:
## Background

JSON mode writes its machine-readable failure to stdout. Automation normally captures stdout for parsing, so a non-zero command could look silent to a human or coding agent even though Alien had an actionable typed error.

Closes ALIEN-884.

## Reproduction

Run an `alien --output json ...` command that returns an API error while capturing stdout. Before this change stderr has no error code, message, or request ID.

## Root cause

The top-level CLI converted and serialized the error only through the selected JSON output renderer. Unlike human mode, there was no terminal diagnostic channel.

## Fix

JSON mode still emits the canonical machine-readable external error on stdout, and now also prints a one-line safe diagnostic on stderr containing only:

- external error code
- sanitized external message
- request ID, when present

Newlines are flattened. Arbitrary error context and internal source details are deliberately excluded.

## Verification

`cargo test -p alien-cli json_error_diagnostic --lib -j 4`

Tests prove request-ID visibility, omission of secret context, and sanitization of internal errors.


alongubkin:
@greptileai review

alongubkin:
@greptileai re-review

PR #660, opened 2026-09-17
fix: recover direct teardown without a local tracker

alongubkin:
## Background

Direct setup is owned by the generated CLI, not Terraform or CloudFormation. The manager persists deployment/setup state, but `destroy` refused to begin unless the original machine's local tracker still existed. An ephemeral runner lost after partial provisioning therefore forced bespoke cloud cleanup.

Closes ALIEN-885.

## Reproduction

1. Begin `alien-deploy deploy` and allow setup-owned cloud resources to be recorded.
2. Lose the original runner and its `~/.config/alien/deployments.json`.
3. On a clean runner, supply the deployment ID, replacement deployment token, manager URL, and normal target-cloud credentials.
4. Before this change, `destroy` has no selector for that deployment and exits before contacting the manager.

## Ownership model

This applies only to direct AWS/GCP/Azure setup. Imported Terraform and CloudFormation deployments remain owned by their IaC state and retain the existing import guard. Local and Machines recovery are explicitly rejected because their host-local lifecycle semantics differ.

## Fix

- Extend the existing `destroy` command with a mutually exclusive `--deployment-id` selector.
- Require explicit supported manager authentication for fresh-runner recovery.
- Fetch the authoritative deployment before cloud mutation and derive platform/import state remotely.
- Reuse the existing idempotent manager cleanup and setup-owned `push_deletion` loop.
- Do not create or remove tracker state for an ID-only invocation.
- Preserve the tracked-name path and remove its tracker entry only after successful cleanup.

Example:

```sh
alien-deploy destroy \
  --deployment-id dep_123 \
  --token-file /run/secrets/deployment-token \
  --manager-url https://manager.example.com
```

## Verification

- `cargo check -p alien-deploy-cli`
- `cargo test -p alien-deploy-cli commands::down::tests --lib`
- 9 focused tests pass, including selector exclusivity and fresh-runner parsing.


alongubkin:
@greptileai review

alongubkin:
@greptileai re-review

PR #663, opened 2026-09-17
feat(sandbox): grant a GCP sandbox read on the registry it pulls from

ItamarZand88:
## Summary

A GCP sandbox runs a container it pulls from the registry Alien hosts, but the
grant machinery counted a sandbox only on AWS, so the image it needs was never
covered. This grants it, and makes the service types the grant and the revoke
name one definition instead of two.

When a deployment reconciles its registry access:

1. The platform decides whether this deployment's sandbox pulls from Alien's
   registry at all. AWS builds a MicroVM image from a published bundle and GCP
   pulls a container, so both earn a grant; Azure names an image from its own
   catalog and earns nothing.
2. **Each compute service is granted as its own Google-managed service agent.**
   A sandbox pulls as a different principal from a worker, so a grant made for
   one reaches neither the other.
3. On teardown the last project consumer revokes the same list, because a member
   the revoke does not name stays on the repository policy with nothing left to
   remove it.

This changes a hardcoded "only AWS sandboxes pull from us" to a question asked
per platform.

## What I did

Replaced the `INCLUDE_SANDBOX` / `EXCLUDE_SANDBOX` constant pair with a function
that answers per platform, and added a sandbox compute-service type mapped to the
Vertex sandbox service agent.

Moved the grant off the deployment's extra service-account emails, which was the
wrong place: those are revoked per deployment, so the first deployment deleted in
a project would strip a binding its siblings still need. The service-type list is
refcounted and survives until the last consumer goes.

Fixed a regression the widened guard introduced. That guard is deliberately
prefix-blind, so it now admits a GCP sandbox whose image Alien never hosted. Such
a deployment was never granted, and on a manager that hosts no images there is no
registry binding to load, so its delete reconcile started returning an error every
tick. It now answers that there is nothing to strand.

## How I tested

`cargo test -p alien-manager -p alien-bindings --lib` — 239 and 338 pass, and
`cargo check --workspace --all-targets` is clean, on a branch rebased onto current
main.

Pinned the new behaviour rather than assuming it: a GCP sandbox-only stack is
covered by the grant, the shared project grant survives while any sibling
deployment is alive, and a never-granted GCP sandbox on a public image reaches
the binding lookup with nothing recorded to clean up.

Checked the attacks this change makes possible:

- A tenant reading another tenant's images. Pre-existing and unchanged by this
  diff: GCP has one registry repository per environment, so the worker agent
  already holds this. Tracked separately; this adds a second principal inside the
  same customer project, which holds no capability the customer lacks today.
- A grant outliving its deployment. The revoke names every service type, and
  removal is a `retain`, so naming one never granted is a no-op while omitting one
  would orphan a member (`gar.rs:207`).
- One deployment's teardown stripping a live sibling's access. The project-level
  members are revoked only in the last-consumer branch, and the per-deployment
  identity stays per-deployment (`registry_access.rs:92-152`).
- Granting to a principal that does not exist yet. The sandbox service agent is
  created at a project's first session, so the binding necessarily precedes it.
  The binding is accepted; whether it becomes effective when the agent appears is
  measured by the cross-project pull test, not claimed here.

Nothing turned up.


ItamarZand88:
Two fixed in 49a4fb29c, one is real and tracked but out of scope here.

**Cleanup Skips Live Grants — fixed, and it was my regression.** You are right, and the comment six lines above the guard already said so: the marker is written after the cloud grant lands, so a false one still covers a live grant. My early return would have left registry read alive after the deployment was deleted. Reverted to failing loudly.

For context on what I was trying to fix: counting a GCP sandbox widened a prefix-blind guard, so a GCP sandbox whose image Alien never hosted now reaches the binding lookup and errors on a manager that hosts no images. That behaviour is pre-existing on AWS, where sandboxes were always counted, so GCP now simply matches it. Erroring loudly beats silently dropping a revoke.

**Sandbox Permissions Decode Incorrectly — fixed.** The read path now derives its service agents from the same table the write path uses, so a member decodes as the type that produced it. Added `every_service_type_survives_a_write_then_read` as the round-trip test you asked for.

**Existing Grants Stay Incomplete — real, tracked, not fixed here.** `registry_access_granted` is a one-shot latch with no `false` assignment anywhere in the tree, so this is broader than the sandbox type: no change to what a grant contains ever reaches an already-granted deployment. The fix is to record what was granted (a member-set fingerprint) rather than that something was, which is a runtime-metadata schema change that wants its own review. It is filed and raised to urgent.

Worth knowing alongside it: this binding only matters cross-project. Google grants the sandbox agent a project-scoped role that already carries `artifactregistry.repositories.downloadArtifacts`, so a same-project pull needs nothing from us. Whether a binding created before the agent exists becomes effective once it materialises is not yet measured, and that is the gating question for this feature rather than for this diff.

ItamarZand88:
@greptile review

ItamarZand88:
Both valid. Neither is fixed here, and here is why.

**Issue 1 — the distinction you are asking for cannot be made at this call site.** You are right that the path should tell an unpersisted partial grant apart from a deployment that could never have been granted. It cannot, because `load_artifact_registry` (`registry_access.rs:661`) reaches `None` through `if let Ok(ar) = ...`, so `None` means both *no binding configured* and *binding failed to load*. The marker does not break the tie either: it is written after the cloud grant lands, so a false one still covers a live grant.

That leaves two wrong answers, and I have now written both. Returning `Ok` on an unset marker was my previous commit, and you correctly called it a dropped revoke. Erroring is what is there now: loud, and wrong only for a deployment that could never have been granted on a manager with no registry binding at all.

I am keeping the loud one. A cleanup that retries is recoverable; a grant that outlives its deployment is not. The real fix is to stop discarding the reason one function up, which changes that signature for every caller, so it is filed separately rather than attempted as a third version of this guard inside a feature PR.

**Issue 2 — real, filed, and broader than this diff.** `registry_access_granted` has no `false` assignment anywhere in the tree, so it is not only the sandbox type that never reaches an already-granted deployment; no change to a grant ever does. The fix is to record what was granted rather than that something was. Filed as urgent.

One thing worth knowing about the blast radius of both: this binding only matters cross-project. Google grants the sandbox agent a project-scoped role that already carries `artifactregistry.repositories.downloadArtifacts`, so a same-project pull needs nothing from us.

ItamarZand88:
Ordering note: this waits on #675, which is at 5/5 and touches the same code in `registry_access.rs`.

#675 gives `collect_image_repositories` / `repository_ids_for_access` a `project_id` argument, binds a sandbox's grant to `{prefix}-{project_id}`, returns a refusal flag that keeps an incomplete grant from being marked done, and reworks the cleanup guard this PR's first finding is about.

Once #675 merges I'll rebase this branch onto it, port the per-platform sandbox question onto the new signatures, and fix the two findings here:
1. a manager with no registry binding failing cleanup forever for a never-granted GCP sandbox on a public image;
2. existing GCP deployments whose grant is already recorded never picking up the sandbox service agent.

ItamarZand88:
Rebased onto main now that #675 has landed, and squashed to one commit.

The two designs had to merge rather than stack: #675 made a sandbox's registry reference `privateBaseImage`, which is AWS-only, while this PR needs a GCP sandbox counted through its own `code.image`. Which field names the image is now a per-platform question (`sandbox_image_source`): AWS reads `privateBaseImage`, GCP reads `code.image`, Azure reads neither. The project-bound grant, the refusal flag and the cleanup fallback from #675 are all kept.

alien: manager 247, bindings 339, `cargo check --workspace --all-targets` clean.

@greptile review

ItamarZand88:
@greptile review

ItamarZand88:
@greptile review

PR #666, opened 2026-09-18
ci: build the package gateway in the unified CLI graph

alongubkin:
## Summary

- build `alien-ai-gateway` in the same development-profile Cargo invocation as all required CLIs
- move that unified build before package-layout validation so the executable consumer test reuses the gateway artifact
- retain the isolated release-profile fallback for local/package-only paths where the unified CLI build is intentionally skipped

## Why

The required TypeScript/packages job previously built the gateway alone in release during package-layout validation, then compiled all CLIs in development. Actions run 34903156597 spent **4m33s** in the isolated gateway release build.

The first candidate changed the fallback itself to development mode. CI correctly rejected that design: the narrower standalone gateway command populated a development-profile `alien-platform-api`, and the later broader CLI command lacked `get_command` generated types. This revision resolves the union once by compiling the gateway and CLIs in a single Cargo invocation.

## Validation

- the original cold development gateway build compiled successfully and the executable reported its expected usage
- the failed first CI run provided a real regression test for cross-command feature reuse
- the revised workflow passes actionlint
- TypeScript formatting and `git diff --check` pass
- the 6.0 GiB local validation target was deleted immediately
- fresh complete TypeScript/packages CI is the authoritative correctness and timing test

Linear: ALIEN-899


alongubkin:
The current CI failure is caused by the merged filtered Platform client regression now fixed in #667, not by this experiment: `main` omits `getCommand` from the operation allowlist, so the ordinary CLI graph lacks `get_command`, `CommandResultAvailability`, and `CommandState`. I reproduced the same failure with a plain fresh `cargo check -p alien-cli` on #652. I’ll rebase this PR after #667 merges and use that clean run for the timing decision.

PR #669, opened 2026-09-18
ci: enable Rust compiler caching

alongubkin:
## Summary

- enable the already-installed sccache binary as Cargo’s rustc wrapper
- preserve uncached fallback when the optional cache setup action fails

## Evidence

The latest required Rust job installed sccache but its post-run report showed exactly 0 compile requests, 0 cache hits, and 0 cache misses after a 17m16s workspace build. `mozilla-actions/sccache-action` installs and configures the cache backend but does not set `RUSTC_WRAPPER` for Rust.

Baseline after #668: 19m17s complete Rust job / 17m16s workspace step. This PR must first show nonzero sccache requests, then a second equivalent run will measure warm-cache impact.

Linear: ALIEN-903


alongubkin:
Closing after the required controlled warm-cache measurement.

Population run:
- workspace step: 24m36s vs 17m16s uncached baseline
- 2,898 writes; cumulative remote write time 751s

Identical warm rerun:
- 2,414 Rust hits / 50 misses (98.0% Rust hit rate)
- workspace step: 17m11s, only 5s / 0.5% faster than baseline
- complete job: 18m54s, only 23s / 2.0% faster than baseline
- cumulative cache-read time: 347s

The remote WebDAV latency consumes nearly the entire compiler-time saving and makes cold/cache-invalidating builds about seven minutes slower. This fails the complete-workflow bar, so we will not enable the same wrapper in Platform.

PR #675, opened 2026-09-19
feat(sandbox): build a Live sandbox from a project's private base image

ItamarZand88:
## Summary

An AWS sandbox can now build from a private ECR base image, and a manager can create a project's image repository on request so there is somewhere to push it.

When a Live sandbox declares `privateBaseImage`:

1. The deployment registers, and the manager learns the customer's AWS account.
2. **The manager grants that account read on the project's own repository, `{prefix}-{projectId}`, and refuses any other repository named there.** ← the heart
3. The runtime builds the image, and its build role pulls the private base image.

This PR changes a sandbox from "public base images only" to "a private base image in its project's repository, on the Live lifecycle".

## What was broken

The cross-account grant read a sandbox's `code.image`, which holds its S3 bundle, not a registry reference. So it never opened any repository, and a private base image could not be pulled.

## What I did

- Added `privateBaseImage` to the sandbox (AWS only). It is the only field naming the repository the build pulls from.
- Bound the grant to the deployment's own project. The repository prefix is shared by every project, so any other name is refused. A refusal keeps the grant marked incomplete, so reconciliation logs it every cycle instead of marking the deployment done.
- Refused a private base image on a Frozen sandbox at plan time. Setup builds a Frozen image before the deployment registers, and only registration tells the registry which account to open, so that build could never pull.
- Added `POST /v1/image-repositories`. ECR accepts a push only into a repository that already exists, while GAR and ACR create one on first push; this route makes the three behave the same.
- Access goes through a new `Authz::can_provision_image_repository`, denied by default. A narrow `ImageRepositoryProvisioner` role lets an embedder authorize this one call. OSS denies that role every project read, and the registry proxy refuses it every pull.

## Files touched

- `alien-core/src/resources/sandbox.rs`: the new field, refused off AWS.
- `alien-preflights/.../frozen_resource_lifecycle.rs`: refuses Frozen + private base image.
- `alien-manager/src/registry_access.rs`: the project-bound grant; `project_id` comes from the deployment record at every caller (`routes/sync.rs`, `transports/manager.rs`, `loops/deployment.rs`).
- `alien-manager/src/routes/image_repositories.rs`: the new route.
- `alien-manager/src/auth/*`, `providers/oss_authz.rs`: the authz method and the role.
- `alien-manager/src/routes/registry_proxy.rs`: the role is refused pulls.

## How I tested

- **Unit tests:** manager 240, core 582, preflights 328. They cover the project-bound grant (own repository granted; sibling and bare-prefix repositories refused; a mixed worker + sandbox stack stays incomplete), the Frozen refusal, and the role's denials. Each commit's tip was tested separately.
- **Route test** (`tests/registry_proxy_test.rs`): a deployment token gets 403, `local` gets 400, and a missing registry binding returns a sanitized 500.
- **Anything I couldn't test:** a real Live build pulling a private ECR base image cross-account. This needs a live AWS run.

- A sandbox naming another project's repository to get it opened: refused, because the grant accepts only `{prefix}-{project_id}` from the deployment record (`registry_access.rs`, test `aws_registry_access_skips_sandbox_images_outside_the_repository_prefix`).
- The provisioner role used to read, push, or pull: every project read is denied in OSS (test `the_image_repository_provisioner_has_no_project_access`), and pulls are refused before the project-scope bypass (test `the_image_repository_provisioner_cannot_pull`).
- An embedder that makes no decision: `can_provision_image_repository` defaults to deny.
Nothing turned up.

## Out of scope

- A private base image on a Frozen sandbox: that needs the grant to land before setup builds the image.
- A worker image naming another project's repository gets the same prefix-only check as before; this PR does not change worker grants.
- A grant is not revoked if a later release drops `privateBaseImage`, because cleanup re-derives repositories from the current stacks.


ItamarZand88:
Both fixed in 9a22c490a.

1. A dropped `privateBaseImage` left a grant nothing could name. Cleanup now derives the project's own repository whenever the deployment still declares a sandbox, so the grant is revoked; test `a_dropped_base_image_still_names_the_repository_cleanup_must_revoke`.
2. The check compared only the repository path. The reference must now name the account the grant is written in, so a same-path repository in another account is refused; covered in `aws_registry_access_skips_sandbox_images_outside_the_repository_prefix`.

@greptile review

ItamarZand88:
Both fixed in a13e03cf2.

1. The region is matched as well now: a reference in the configured account but another region is refused, since the grant would open the repository in one region while the build pulls from another. A stored reference's `{region}` token still matches, because it stands for whichever region the deployment renders.
2. The cleanup fallback now names the project's repository only when the registry actually has it, so a worker-only grant on the shared repository no longer sends cleanup at a repository nobody created.

@greptile review

ItamarZand88:
Fixed in the latest commit. Only a not-found answer counts as absent now; any other lookup failure fails the cleanup, so the grant is retried rather than silently left in place.

@greptile review

PR #680, opened 2026-09-21
fix: prepare composed CLI release dependencies

alongubkin:
## Summary
- check out the private Horizon dependency before composed CLI builds
- pin the official composed CLI Platform revision to the merged helper action
- apply the same setup on Linux, Linux ARM64, macOS, and Windows

## Validation
- release workflow YAML parses successfully
- `git diff --check` passes
- Platform helper action was independently reviewed and merged with a 5/5 Greptile score

Linear: ALIEN-912


alongubkin:
Addressed the routing-coverage finding in 45e525035: the existing release-routing test now verifies all four composed CLI jobs run the dependency setup after the pinned Platform checkout and pass the repository token. The focused test passes.

@greptileai review

alongubkin:
Tightened the assertion in fa964b5e4 so the token must be inside the dependency action’s own `with` block; it can no longer match a later setup step. Focused test remains green.

@greptileai review

alongubkin:
Admin merge rationale: Greptile 5/5 with the prior finding fully addressed; release qualification, PR conventions, Enclave, and the focused routing test are green. The remaining broad TypeScript job is still in environment setup and the change is limited to release workflow wiring plus its regression assertion.

